PR #274, opened 2017-12-09
fix: store should share the same trielock

0x04C2:

When store tries to s.Bucket("otherBucket"), the otherBucket shares the same trie data, but the lock is different. For now, store is only used to get `meta.json`. For safety reason, we should share the trielock.

NONE

**3.Describe how you did it**
NONE

NONE

The race test is hard to do in the UT. @allencloud


skoo87:
Thanks, I think that you are right. @0x04C2

PR #327, opened 2017-12-14
fix: make 'pouch ps' list containers orderly and show creation time.

zeppp:

make `pouch ps` list containers ordered by its creating time instead randomly and show creation time.

**3.Describe how you did it**

```
pouch ps
Name   ID       Status    Image                              Runtime   Created
2      f93a25   created   docker.io/library/busybox:latest   runc      about 1 hour ago
1      4ddedc   created   docker.io/library/busybox:latest   runc      about 1 hour ago

pouch create --name 3 docker.io/library/busybox:latest 
container ID: bd618e450eb9f25d101211b6c306c2292b4a01402dbef7f249cc53a9fbc921d2, name: 3 

pouch ps
Name   ID       Status    Image                              Runtime   Created
3      bd618e   created   docker.io/library/busybox:latest   runc      about 4 seconds ago
2      f93a25   created   docker.io/library/busybox:latest   runc      about 1 hour ago
1      4ddedc   created   docker.io/library/busybox:latest   runc      about 1 hour ago
```
Waiting 2 minutes
```
pouch ps
Name   ID       Status    Image                              Runtime   Created
3      bd618e   created   docker.io/library/busybox:latest   runc      about 2 minutes ago
2      f93a25   created   docker.io/library/busybox:latest   runc      about 1 hour ago
1      4ddedc   created   docker.io/library/busybox:latest   runc      about 1 hour ago

```

### DIFF
--- a/cli/ps.go
+++ b/cli/ps.go
@@ -2,12 +2,19 @@ package main
 
 import (
 	"fmt"
+	"sort"
+	"strconv"
+	"time"
 
+	"github.com/alibaba/pouch/apis/types"
 	"github.com/spf13/cobra"
 )
 
 // psDescription is used to describe ps command in detail and auto generate command doc.
-var psDescription = "\nList Containers with container name, ID, status, image reference and runtime."
+var psDescription = "\nList Containers with container name, ID, status, image reference, runtime and creation time."
+
+// containerList is used to save the container list.
+type containerList []*types.Container
 
 // PsCommand is used to implement 'ps' command.
 type PsCommand struct {
@@ -39,15 +46,18 @@ func (p *PsCommand) addFlags() {
 func (p *PsCommand) runPs(args []string) error {
 	apiClient := p.cli.Client()
 
+	var containers containerList
 	containers, err := apiClient.ContainerList()
 	if err != nil {
 		return fmt.Errorf("failed to get container list: %v", err)
 	}
 
+	sort.Sort(containers)
+
 	display := p.cli.NewTableDisplay()
-	display.AddRow([]string{"Name", "ID", "Status", "Image", "Runtime"})
+	display.AddRow([]string{"Name", "ID", "Status", "Image", "Runtime", "Created"})
 	for _, c := range containers {
-		display.AddRow([]string{c.Names[0], c.ID[:6], c.Status, c.Image, c.HostConfig.Runtime})
+		display.AddRow([]string{c.Names[0], c.ID[:6], c.Status, c.Image, c.HostConfig.Runtime, createdTime(c.Created)})
 	}
 	display.Flush()
 	return nil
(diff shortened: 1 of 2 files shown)


allencloud:
Please add more details in `4.Describe how to verify it` about your command execution output. 

> We encourage contributors to follow the PR template to finish the pull request.

zeppp:
There is still a problem that the order will be determined by  recover order not creating order  when pouch recovers the containers. But i think it is acceptable. WDYT @allencloud

allencloud:
> There is still a problem that the order will be determined by recover order not creating order when pouch recovers the containers. But i think it is acceptable

Currently I think it is acceptable. If we change the logic of recover, we need to take this into consideration at that time.

janynedoe48:
Pouch-ps.
Sign me out.

janynedoe48:
Do whatever you need to do to stop bothering me please

zeppp:
>  Here is my reason you can check that we have not put creation time in pouch ps. How about first adding creation time in `pouch ps`  output?

It is meaningful for users to know when the container created. I may try to add created time in output.
But I still think  listing container by creation time is meaningful.If an user run a container,he or she must wish to use `pouch ps`  to check the container state at once.If the order is random，he or she may take some time to find the container especially when there are many containers. @allencloud

allencloud:
> But I still think listing container by creation time is meaningful

I agree with you. 

Without showing creation time, this order means less. So maybe we could improve the priority of adding creation time in `pouch ps`.

zeppp:
I have added the creation in `pouch ps`. 
And , the problem still exists.
>  There is still a problem that the order will be determined by recover order not creating order when > 
 pouch recovers the containers.

 I can fix it by sorting the safelist without changing the logic of recover. WDYT @allencloud

allencloud:
> There is still a problem that the order will be determined by recover order not creating order when 
pouch recovers the containers.

If that, @skoo87 please take a look of this and dertermine that if we can fix this.

@zeppp I think it is good for you to add creation time. Fixing recover order could be a follow-up PR.

skoo87:
it's a good job

skoo87:
What is the reason for needing a "safelist" ? Why don't sort it by createtime directly ? @zeppp

zeppp:
>I think We add a new state (a list) for this job in pouchd, it's not a good idea.

You are right , i will fix it later. Thanks for your advice. @skoo87

zeppp:
>There is still a problem that the order will be determined by recover order not creating order when
pouch recovers the containers.

>Why don't sort it by createtime directly in "cli" 

I have fixed it. PTAL @allencloud @skoo87

skoo87:
LGTM

cli/ps.go
```diff
@@ -56,7 +66,63 @@ func (p *PsCommand) runPs(args []string) error {
...
-1dad17   1dad17   stopped   docker.io/library/busybox:latest   runv
-505571   505571   stopped   docker.io/library/busybox:latest   runc`
+Name     ID       Status    Image                              Runtime   Created
+1dad17   1dad17   stopped   docker.io/library/busybox:latest   runv      about 1 hour ago
+505571   505571   stopped   docker.io/library/busybox:latest   runc      about 1 hour ago
+`
+}
+
+// createdTime is used to show the time from creation to now.
+func createdTime(created string) (s string) {
```
allencloud: please move this function to pkg, since it could be a common function which is used other places.
And add more unit-test for this function in pkg.

PR #515, opened 2018-01-05
bugfix: exec should return 200 if client doesn't upgrade proto

0x04C2:

If user calls `/exec/execid/start` without upgrade header, the pouchd should return 200 status.
But now, the pouchd always return 101 switch protocol. This PR is to fix this bug.

fixes #494 

**3.Describe how you did it**

For the AttachConfig, the `Upgrade` field should check header from request.

I have added the test for this.

In this PR, I upgrades the `request.Hijack` function and add more comments here: 

For hijack request, the server can return 200 status or 101 switch proto. 

If http-client doesn't switch proto, the `Do` function will return error caused by `ErrPersistEOF`.

As the RFC 2616, section 4.4 mentioned,  if there is no Content-Length or chunked Transfer-Encoding on a *Response and the status is not 1xx, 204 or 304, then the body is unbounded. And the response is always terminated by the first empty line after the header fields.

In our case, the 200 response is like:

```
HTTP/1.1 200 OK
Content-Type: application/vnd.docker.raw-stream
<= first empty line
[raw-data]
```

The golang http client does follow the RFC 2616. More detail is here:
- https://github.com/golang/go/blob/release-branch.go1.9/src/net/http/response.go#L190
- https://github.com/golang/go/blob/release-branch.go1.9/src/net/http/transfer.go#L486:L518

For this case, both status and header has been read into response. And the [raw-data] will be read from hijack.reader. So we can ignore the `ErrPersistEOF` here.

cc @allencloud  and @Letty5411

### DIFF
--- a/apis/server/container_bridge.go
+++ b/apis/server/container_bridge.go
@@ -83,6 +83,7 @@ func (s *Server) startContainerExec(ctx context.Context, rw http.ResponseWriter,
 	}
 
 	name := mux.Vars(req)["name"]
+	_, upgrade := req.Header["Upgrade"]
 
 	var attach *mgr.AttachConfig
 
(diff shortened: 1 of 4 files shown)


Letty5411:
LGTM

PR #814, opened 2018-03-05
fix: optimize the way of packaging

zzchun:
Signed-off-by: Zhanchun Zhang <zzchun@zju.edu.cn>

Optimize the way of packaging

NONE

### DIFF
--- a/hack/package/deb/build.sh
+++ b/hack/package/deb/build.sh
@@ -1,39 +1,20 @@
 #!/usr/bin/env sh
 
 set -e
-# This script is to build pouch rpm package as follows,
-# Following the below command to build rpm
-# 1. Build pouch:deb image
-#	cd pouch
-#	docker build -t pouch:deb -f hack/package/deb/ubuntu-xenial/Dockerfile.x86_64 .
-# 2. Mount a directory which contains gpg keys, eg
-#    $ tree /root/deb/
-#		deb
-#		├── keys
-#		│   ├── gpg
-#		│   └── secretkey
-# 3. run pouch:deb to generate deb package
-#	 docker run -it -e VERSION="0.1.0" -v /root/deb/:/root/deb pouch:deb bash -c hack/package/deb/build.sh
-# 4. In this example, rpm package will be output in '/root/deb/package/' directory
-
-DIR="$( cd "$( dirname "$0" )" && pwd )"
 
 TMP=$(mktemp -d /tmp/pouch.XXXXXX)
-
+DEFAULT_GPG_KEY=${GPGKEY:-"439AE9EC"}
 MOUNTDIR=/root/deb
-PACKAGEDIR=/root/deb/package
+PACKAGEDIR=/root/deb/package/deb/
 BASEDIR=/go/src/github.com/alibaba/pouch
 
-rm -rf $BASEDIR/debian && cp -r $DIR/debian $BASEDIR/
-rm -rf $BASEDIR/systemd && cp -r $DIR/systemd $BASEDIR/
-
-if [ ! -e $$BASEDIR/Makefile.bak ] ;then
-	mv $BASEDIR/Makefile $BASEDIR/Makefile.bak
-	cp $DIR/Makefile $BASEDIR/Makefile
-fi
+# prepare files that used to build deb packages
+rm -rf $BASEDIR/debian && cp -r $BASEDIR/hack/package/deb/debian $BASEDIR/
+rm -rf $BASEDIR/systemd && cp -r $BASEDIR/hack/package/deb/systemd $BASEDIR/
+rm -f $BASEDIR/Makefile.bak && mv $BASEDIR/Makefile $BASEDIR/Makefile.bak
+cp $BASEDIR/hack/package/deb/Makefile $BASEDIR/Makefile
 
 cd $BASEDIR
-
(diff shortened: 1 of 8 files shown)


hack/package/deb/build.sh
```diff
@@ -1,35 +1,19 @@
...
 
-rm -rf $BASEDIR/debian && cp -r $DIR/debian $BASEDIR/
-rm -rf $BASEDIR/systemd && cp -r $DIR/systemd $BASEDIR/
+rm -rf $BASEDIR/debian && cp -r $BASEDIR/hack/package/deb/debian $BASEDIR/
+rm -rf $BASEDIR/systemd && cp -r $BASEDIR/hack/package/deb/systemd $BASEDIR/
 
 if [ ! -e $$BASEDIR/Makefile.bak ] ;then
 	mv $BASEDIR/Makefile $BASEDIR/Makefile.bak
-	cp $DIR/Makefile $BASEDIR/Makefile
+	cp $BASEDIR/hack/package/deb/Makefile $BASEDIR/Makefile
```
Letty5411: Maybe the `cp` option could put outside `if` scope, as even if there is a Makefile.bak, it doesn't mean the Makefile is which we want.

hack/package/deb/build.sh
```diff
@@ -63,3 +47,5 @@ apt-ftparchive release ./ > Release
 # if you want to use your gpg key, you should replace '439AE9EC' with your key
 gpg -abs --default-key 439AE9EC -o Release.gpg Release
 gpg --clearsign --default-key 439AE9EC -o InRelease Release
+
```
Letty5411: Could we use a variable like DEFAULT-GPG-KEY=${GPGKEY:-"439AE9EC"} to make gpg key configurable.

hack/package/deb/build.sh
```diff
@@ -63,3 +47,5 @@ apt-ftparchive release ./ > Release
...
+echo "deb package was output to $PACKAGEDIR"
```
Letty5411: How about:
echo "Build deb package successfully ! Please get the package in $PACKAGEDIR !"

hack/package/package.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# VERSION the version to give to the package
+VERSION=$1
```
Letty5411: use VERSION=${1:?"Version must required"}  to give error message is the package.sh script is not used correctly. The same for the following two variables.

hack/package/package.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Replace Version, BuildTime and CommitID in version.go
+pushd $DIR
+sed -i "s#^const Version.*#const Version = \"$VERSION\"#g" version/version.go
+sed -i "s#^var BuildTime.*#var BuildTime = \"$BUILDTIME\"#g" version/version.go
+sed -i "s#^var GitCommit.*#var GitCommit = \"$GITCOMMIT\"#g" version/version.go
+
+# build images
+docker build -t pouch:rpm -f $DIR/hack/package/rpm/centos-7/Dockerfile.x86_64 .
+docker build -t pouch:deb -f $DIR/hack/package/deb/ubuntu-xenial/Dockerfile.x86_64 .
```
Letty5411: add error check here as docker build is likely to fail.
e.g. 
```
(( $? != 0 )) && echo "build image failed" && exit 1
```

hack/package/package.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# build rpm packages
+docker run -it \
+	-e VERSION="$VERSION" \
+	-e ITERATION="$ITERATION" \
+	-v $KEYDIR:/root/rpm \
+	pouch:rpm
+
```
Letty5411: `docker run -it` change to `docker run --rm`
check the exit code too.

hack/package/package.sh
```diff
@@ -0,0 +1,44 @@
...
+# VERSION the version to give to the package
+VERSION=$1
+
+# The iteration to give to the package. RPM calls this the 'release'.
+# FreeBSD calls it 'PORTREVISION'. Debian calls this 'debian_revision', eg:
+ITERATION=$2
+
+# the gpg keys directory
+KEYDIR=$3
+
```
Letty5411: Add an option to choose build deb/rpm package or both. for example:
```
PKGTYPE=${4:-"all"}
```
And change the following build code like:
```
if [[ $PKGTYPE == "all" || $PKGTYPE == "deb" ]] ; then 
   # build deb image
   # build deb package
fi
if [[ $PKGTYPE == "all" || $PKGTYPE == "rpm" ]] ; then 
  #build rpm image
   # build rpm package
fi

```

hack/package/package.sh
```diff
@@ -0,0 +1,68 @@
...
+		build_rpm
+	elif [[ $PKGTYPE == "deb" ]]; then
+		build_deb
+	else
+		build_rpm
+		build_deb
+	fi
+}
+
+main "$@"
```
Letty5411: no need to use "$@", as main function doesn't handle the input args.

Letty5411:
@zzchun Could you also add an build example in the PR description?

Letty5411:
LGTM

PR #835, opened 2018-03-07
bugfix: use host network when build package

Letty5411:

Use host network when use docker, in case there is networking error.


sunyuan3:
LGTM

PR #885, opened 2018-03-13
refactor: set lxcfs managed by systemd

CodeJuan:
To improve availability, lxcfs process should be managed by other process, such as systemd.

NONE

1. Remove the code of starting lxcfs in Pouchd. 
2. Check if the mount point of lxcfs is existed when starting Pouchd.

1. Make sure lxcfs is running, and the mount point of lxcfs is `/var/lib/lxcfs`
2. `pouchd --enable-lxcfs`
3. `pouch run -m 512M --enableLxcfs=true  registry.hub.docker.com/library/busybox:latest head -n 1 /proc/meminfo`
4. The output should be `524288 kB`

### DIFF
--- a/hack/package/rpm/service/lxcfs.service
+++ b/hack/package/rpm/service/lxcfs.service
@@ -10,6 +10,7 @@ KillMode=process
 Restart=on-failure
 ExecStopPost=-/usr/bin/fusermount -u /var/lib/lxcfs
 Delegate=yes
+ExecStartPost=pouch remount-lxcfs
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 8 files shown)


Letty5411:
`hack/make.sh` also need to be edited to install fuse.

hack/package/rpm/service/lxcfs.service
```diff
@@ -10,6 +10,7 @@ KillMode=process
 Restart=on-failure
 ExecStopPost=-/usr/bin/fusermount -u /var/lib/lxcfs
 Delegate=yes
+ExecStartPost=pouch remount-lxcfs
```
Letty5411: also need to change `./package/deb/systemd/pouch.service`

HusterWan:
LGTM

xigang:
The same problem.
```
# cat /var/lib/lxc/lxcfs/proc/cpuinfo
cat: /var/lib/lxc/lxcfs/proc/cpuinfo: Transport endpoint is not connected
```
@CodeJuan @Letty5411

xigang:
https://github.com/xigang/lxcfs-admission-webhook/blob/dev/script/container_remount_lxcfs.sh

Issue #924, opened 2018-03-20
[bug] pull images failure impacts `pouch images`


Letty5411:
### Ⅰ. Issue Description
Pull image failure makes `pouch images` doesn't work:
```
[root@e19h19470.et15sqa /root]
#pouch pull registry.hub.docker.com/library/httpd:latest
registry.hub.docker.com/library/httpd:latest:                                  resolved       |++++++++++++++++++++++++++++++++++++++|
index-sha256:bd13a0fc2aa205653032b3fef7895332bbbbc6fc9b0552e42d4b33b4a5481290: done           |++++++++++++++++++++++++++++++++++++++|
elapsed: 4.7 s                                                                 total:  1.1 Ki (229.0 B/s)
Error: failed to display progress: failed to pull image: manifest sha256:bd13a0fc2aa205653032b3fef7895332bbbbc6fc9b0552e42d4b33b4a5481290: not found

[root@e19h19470.et15sqa /root]
#pouch images
Error: failed to get image list: {"message":"manifest sha256:bd13a0fc2aa205653032b3fef7895332bbbbc6fc9b0552e42d4b33b4a5481290: not found"}
```

Using Containerd list image is OK:
```
root@osboxes:pouch (master) -> ctr --address /var/run/containerd.sock images ls
REF                                                TYPE                                                      DIGEST                                                                  SIZE      PLATFORMS                                                                                              LABELS
registry.hub.docker.com/library/busybox:latest     application/vnd.docker.distribution.manifest.list.v2+json sha256:2107a35b58593c58ec5f4e8f2c4a70d195321078aebfadfbfb223a2ff4a4ed21 710.8 KiB linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/ppc64le,linux/s390x  -
registry.hub.docker.com/library/hello-world:latest application/vnd.docker.distribution.manifest.list.v2+json sha256:97ce6fa4b6cdc0790cda65fe7290b74cfebd9fa0c9b8c38e979330d547d22ce1 6.0 KiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v7,linux/arm64/v8,linux/ppc64le,linux/s390x,windows/amd64 -
[...]

PR #925, opened 2018-03-20
test: quick fix to skip pull httpd image

Letty5411:

Due to issue #924 , the latest httpd image has problem which impact other tests.
In order to make CI works, skip this test. 
And use a stable tag later in test.


HusterWan:
LGTM

PR #946, opened 2018-03-22
bugfix: fix check kernel version is wrong

rudyfly:
check kernel version when setting disk quota driver, it should check
kernel > 4, instead of major > 4.

For example，the version of kernel is 3.10, and it supports group quota. you can test like this:
```
# cd pouch/test
# go test -check.f TestAliKernelDiskQuotaWorks
```
test will be succeed.


Letty5411:
LGTM

PR #956, opened 2018-03-24
bugfix: initialize the network log

rudyfly:
revert commit: 8a70a80
libnetwork's logrus version is different from pouchd, so we need to
set libnetwork's logrus addintionly.

run pouchd and see network log format

also cc @faycheng 


allencloud:
LGTM

u2takey:
@rudyfly @allencloud may be not easy, because import path not match with actual git path.
Aan we just upgrade libnetwork, which seems fixed this https://github.com/docker/libnetwork/pull/1856/files. 
Or fix makefile so that users can build pouch smoothly.

PR #1045, opened 2018-04-03
bugfix: distinguish cmd and entrypoint better

YaoZengzeng:

In order to be consistent with Docker:

1. User has specified container's entrypoint, don't use the cmd of image's configuration to fill it , even if it is empty.

2. User has not specified container's entrypoing, use the entrypoint of image's configuration to fill it, as well as cmd.

### DIFF
--- a/daemon/mgr/container_types.go
+++ b/daemon/mgr/container_types.go
@@ -171,11 +171,13 @@ func (meta *ContainerMeta) merge(getconfig func() (v1.ImageConfig, error)) error
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


YaoZengzeng:
@allencloud PTAL

With this PR being merged, the node-e2e-test result of k8s + pouch will be:

Ran 186 of 258 Specs in 1673.546 seconds
FAIL! -- 181 Passed | 5 Failed | 0 Pending | 72 Skipped

HusterWan:
LGTM

PR #1106, opened 2018-04-11
test:add container create test

Dewey-Ding:
Signed-off-by: Dewey-Ding <deweyding@gmail.com>

add container create test

### DIFF
--- a/client/container_create_test.go
+++ b/client/container_create_test.go
@@ -1,0 +1,69 @@
+package client
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/alibaba/pouch/apis/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestContainerCreateError(t *testing.T) {
+	client := &APIClient{
+		HTTPCli: newMockClient(errorMockResponse(http.StatusInternalServerError, "Server error")),
+	}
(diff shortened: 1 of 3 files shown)


client/container_create_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+func TestContianerCreate(t *testing.T) {
+	expectedURL := "/containers/create"
+
+	httpClient := newMockClient(func(req *http.Request) (*http.Response, error) {
+		if !strings.HasPrefix(req.URL.Path, expectedURL) {
+			return nil, fmt.Errorf("expected URL '%s', got '%s'", expectedURL, req.URL)
+		}
```
ZouRui89: I think this test can be enhanced. Would you check the type of the request or the data type that the request method gets, which can improve the degree of accuracy greatly. Thx a lot!

client/container_create_test.go
```diff
@@ -0,0 +1,62 @@
...
+		HTTPCli: httpClient,
+	}
+
+	container, err := client.ContainerCreate(context.Background(), types.ContainerConfig{}, &types.HostConfig{}, &types.NetworkingConfig{}, "container_name")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if container.ID != "container_id" {
+		t.Fatalf("expected `container_id`, got %s", container.ID)
+	}
```
ZouRui89: You can assert the value of both id and name here, instead of just that of id.

ZouRui89:
You can refer to similar mock test, such as 'network create' and 'volume create'.

ZouRui89:
Well done! LGTM

PR #1111, opened 2018-04-11
[feature] setup profiler and don't bother to enable debug level log

yyb196:

In production environment i need to enable profilers like `pprof` `stack dumping` and `gops` always for the emergency cases or odd cases, in these cases i need to attach to the running pouchd to get the `cpu` `memory` or `goroutine stack` snapshot online, and then analyze why the odd things happen,  but in these case we don't want alway print debug level log, which is too much output all the time.

This pr enables users to setup profilers and do not bother to always open debug log output. and if debug log level is specified, it's both enable debug level and setup profilers.


yyb196:
@allencloud @HusterWan please help to review this pr

PR #1126, opened 2018-04-13
make TLS config params can be setup in the file

yyb196:

make TLS config params can be setup in the config file which introduce by falg `--config-file`

@Ace-Tang


Ace-Tang:
LGTM.
```
Since encoding/json cannot do the trick with ">some>deep>childnode" like encoding/xml can do. Nested structs is the way to go.
```
I want to re-implement daemon config merge function.

PR #1129, opened 2018-04-15
feature: gc unused exec processes

Ace-Tang:
add gc to clean unused exec processes

add gc to clean unused exec processes, gc process will be triggered every 5 minutes.

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -150,6 +150,8 @@ func NewContainerManager(ctx context.Context, store *meta.Store, cli ctrd.APICli
 	mgr.Client.SetExitHooks(mgr.exitedAndRelease)
 	mgr.Client.SetExecExitHooks(mgr.execExitedAndRelease)
 
+	go mgr.execProcessGC()
+
 	return mgr, mgr.Restore(ctx)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/collect/safe_map.go
```diff
@@ -25,6 +25,19 @@ func (m *SafeMap) Get(k string) *Value {
 	return &Value{v, ok}
 }
 
+// Values returns all key-values stored in map
+func (m *SafeMap) Values() map[string]interface{} {
```
allencloud: If m.inner equals to nil, do we need to return nil instead of `map[string]interface{}`?

Just a little confused on this.

Ace-Tang: m.inner can not be nil,  since
```
func NewSafeMap() *SafeMap {
 	return &SafeMap{
 		inner: make(map[string]interface{}, 16),
 	}
 }
```

daemon/mgr/container.go
```diff
@@ -1564,6 +1566,35 @@ func (mgr *ContainerManager) setBaseFS(ctx context.Context, meta *ContainerMeta,
 	meta.BaseFS = filepath.Join(mgr.Config.HomeDir, "containerd/state", "io.containerd.runtime.v1.linux", namespaces.Default, info.Name, "rootfs")
 }
 
+// execProcessGC cleans unused exec processes config every 5 minutes.
+func (mgr *ContainerManager) execProcessGC() {
+	for range time.Tick(5 * time.Minute) {
```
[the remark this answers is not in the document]
Ace-Tang: Actually, I do not think time clock need to be a variable.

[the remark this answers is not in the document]
allencloud: I like this feature very much, and it indeed brings much benefit to improve memory efficiency. While I totally agree with the initial point. While to be honest, I am not on a tiny point of this part. How about making this `5 * time.Minute` to be a const variable of `DefaultExecGCInterval`? 

/cc @rudyfly @Letty5411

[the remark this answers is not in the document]
Ace-Tang: I have already update that, but forgot to push. @allencloud

pkg/collect/safe_map.go
allencloud: Someone has some potential possibility that they do not use NewSafeMap() to create a SafeMap. If they create it via `safetyMap := &SafeMap{}`, it works finely. But, if they call function `safetyMap.Values()`, it could panic. Right?

Ace-Tang:
updated, and add more test for collect pkg, @allencloud

pkg/collect/safe_map.go
Ace-Tang: The same reason with #1132 .
for range a nil map will never cause panic, since nil value will convert to a nil map(map[] ), that add a nil check is useless.

allencloud:
Thanks a lot for your update. LGTM

PR #1138, opened 2018-04-17
bugfix: fix flags name should be same

Ace-Tang:
flag name and json field in daemon file should be consistent,
cause we need check the valid and conflict in flags from two part.

flag name and json field in daemon file should be consistent,
cause we need check the valid and conflict in flags from two part.

for eg, this two part should have same name.
```
Labels []string `json:"label,omitempty"`

flagSet.StringSliceVar(&cfg.Labels, "label", []string{}, "Set metadata for Pouch daemon")
```


Ace-Tang:
this pr has conflict with #1132 , if one is merged, the other should be rebased.

Ace-Tang:
it can be merged, @allencloud ,PTAL

HusterWan:
LGTM

PR #1165, opened 2018-04-19
bugfix: persistent storage of sandbox metadata

### LINKED ISSUE #1164: [bug] The pod status data which get from CRI  is error
### Ⅰ. Issue Description
After pod created, if i restart pouch daemon, the sanboxMeta data store in SandboxStore will lost. Now if i get pod status from cri ,it will always get error.

### Ⅱ. Describe what happened

image

pouch version
GitCommit:       unknown
GoVersion:       go1.9.1
KernelVersion:   3.10.0-327.ali2012.alios7.x86_64
Os:              linux
Version:         0.4.0-dev
APIVersion:      1.24
Arch:            amd64
BuildTime:       unknown

YaoZengzeng:

fixes #1164

### DIFF
--- a/daemon/mgr/cri.go
+++ b/daemon/mgr/cri.go
@@ -6,13 +6,14 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"time"
 
 	apitypes "github.com/alibaba/pouch/apis/types"
 	"github.com/alibaba/pouch/cri/stream"
 	"github.com/alibaba/pouch/daemon/config"
-	"github.com/alibaba/pouch/pkg/collect"
+	"github.com/alibaba/pouch/pkg/meta"
 	"github.com/alibaba/pouch/pkg/reference"
 	"github.com/alibaba/pouch/version"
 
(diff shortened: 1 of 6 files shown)


daemon/mgr/cri.go
```diff
@@ -312,7 +327,10 @@ func (c *CriManager) RemovePodSandbox(ctx context.Context, r *runtime.RemovePodS
 		return nil, fmt.Errorf("failed to remove root directory %q: %v", sandboxRootDir, err)
 	}
 
-	c.SandboxStore.Remove(podSandboxID)
+	err = c.SandboxStore.Remove(podSandboxID)
+	if err != nil {
```
HusterWan: Should we return err info to Caller, if we do not deal with the error, there will have more and more dirty data, WDYT?

YaoZengzeng:
@allencloud PTAL

daemon/mgr/cri.go
YaoZengzeng: Agree :)

allencloud:
LGTM

PR #1226, opened 2018-04-26
bugfix: align with docker v1.24 API

### LINKED ISSUE #1221: [bug] exec is not compitable with moby
### Ⅰ. Issue Description
#./docker -H unix:///run/pouchd.sock exec -i 42ed sh -c 'echo test | telnet 11.160.218.230 22'
Unrecognized input header: 84

#pouch exec -i 42ed sh -c 'echo test | telnet 11.160.218.230 22'
Trying 11.160.218.230...
Connected to 11.160.218.230.
Escape character is '^]'.
Connection closed by foreign host.

1.
2.
3.

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

fuweid:
The docker exec api uses custom stream bytes if the tty is false.
In order to align with docker API, we need to use
github.com/docker/docker/pkg/stdcopy package to modify stream bytes.

Align with docker v1.24 API. As we known, the docker API has custom stream bytes when the client doesn't set tty. More information is here: https://github.com/moby/moby/blob/master/client/container_attach.go#L10-L33

Fix #1221

use `github.com/docker/docker/pkg/stdcopy` to wrap the `io.Writer`. 

```
➜  docker ./docker -H unix:///var/run/pouchd.sock exec -i 05e7c1b71cf63137fbac112f5ab2c5a6d9002c938aca304e259507b9c4aec66c ls
Unrecognized input header: 98
```

after fix:

```
➜  docker ./docker -H unix:///var/run/pouchd.sock exec -i 05e7c1b71cf63137fbac112f5ab2c5a6d9002c938aca304e259507b9c4aec66c ls
bin
dev
etc
home
proc
root
run
sys
tmp
usr
var
```

@allencloud @HusterWan @yyb196 This change is not ideal. We need to refactor the IO related things.

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"github.com/alibaba/pouch/apis/types"
+	"github.com/docker/docker/pkg/stdcopy"
 
 	"github.com/spf13/cobra"
 )
@@ -62,7 +63,7 @@ func (e *ExecCommand) runExec(args []string) error {
 
 	createExecConfig := &types.ExecCreateConfig{
 		Cmd:          command,
-		Tty:          e.Terminal || e.Interactive,
+		Tty:          e.Terminal,
 		Detach:       e.Detach,
 		AttachStderr: !e.Detach,
 		AttachStdout: !e.Detach,
@@ -79,7 +80,7 @@ func (e *ExecCommand) runExec(args []string) error {
 	// start exec process.
 	startExecConfig := &types.ExecStartConfig{
 		Detach: e.Detach,
-		Tty:    e.Terminal && e.Interactive,
+		Tty:    e.Terminal,
 	}
 
 	conn, reader, err := apiClient.ContainerStartExec(ctx, createResp.ID, startExecConfig)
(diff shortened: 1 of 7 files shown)


vendor/vendor.json
```diff
@@ -493,6 +493,12 @@
 			"revision": "53a58da551e961b3710bbbdfabbc162c3f5f30f6",
 			"revisionTime": "2018-01-31T23:13:00Z"
 		},
+		{
+			"checksumSHA1": "H1rrbVmeE1z2TnkF7tSrfh+qUOY=",
```
HusterWan: Should we let docker vendor pkg version be same ? :)

fuweid: @HusterWan Yes. It's already "53a58da551e961b3710bbbdfabbc162c3f5f30f6"

cli/exec.go
```diff
@@ -95,19 +96,26 @@ func (e *ExecCommand) runExec(args []string) error {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			io.Copy(os.Stdout, reader)
+			if !e.Terminal {
```
HusterWan: I think we should add some test case for `exec` command  `-i` and `-it` flags

we can add the test case later, because this feature is needed very hurry

fuweid: sure thing.

HusterWan:
LGTM

PR #1233, opened 2018-04-27
bugfix: setRawMode can only be set when the user set tty

fuweid:

`setRawMode` is used to terminal case. If the shell is non-interactive or non-terminal, we should not set the raw mode for the input.

None

check tty before set the raw mode

we need to set the shell in no-interactive mode, please use the following the script:
```
#!/bin/bash

cat >test <<'EOF'
set +i # set non-interactive
pouch run -i busybox:latest echo test
EOF

echo 'running bash -i <test'
bash -i <test
```

before this change:
```
➜  /tmp sh ./test-interactive
running bash -i <test
root@ubuntu-xenial:/tmp# set +i # set non-interactive
root@ubuntu-xenial:/tmp# pouch run -i busybox:latest echo test
Error: failed to set raw mode
root@ubuntu-xenial:/tmp# exit
```

after this change:
```
➜  /tmp sh ./test-interactive
running bash -i <test
root@ubuntu-xenial:/tmp# set +i # set non-interactive
root@ubuntu-xenial:/tmp# pouch run -i busybox:latest echo test
test
root@ubuntu-xenial:/tmp# exit
```

### DIFF
--- a/cli/run.go
+++ b/cli/run.go
@@ -100,15 +100,17 @@ func (rc *RunCommand) runRun(args []string) error {
 	wait := make(chan struct{})
 
 	if rc.attach || rc.stdin {
-		in, out, err := setRawMode(rc.stdin, false)
-		if err != nil {
-			return fmt.Errorf("failed to set raw mode")
-		}
-		defer func() {
-			if err := restoreMode(in, out); err != nil {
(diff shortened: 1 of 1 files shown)


HusterWan:
LGTM

Issue #1243, opened 2018-04-27
[proposal] add vagrant envrionment 


u2takey:
### Ⅰ. Issue Description
As many developer use macos/windows for development, i propose adding vagrant environment  to facilitate development for such users.

If this seems reasonable, i can help to init this environment (vagrantfile as related docs).

HusterWan:
@u2takey thanks a lot for your feedback, maybe @fuweid can research setup the vagrant environment  😀

fuweid:
Hi @u2takey ,  happy to hear!!!

I think it's reasonable. You can show your PR 😀 . If you want to help us to do that, please consider to make the setup script reusable, because the setup way of Vagrant is the same to Dockerfile. 

Please go ahead!

fuweid:
Thanks @u2takey 's help.

PR #1245, opened 2018-04-27
feature: add vagrant environment for development

u2takey:

add vagrant environment for development

https://github.com/alibaba/pouch/issues/1243

### DIFF
--- a/hack/vagrant/README.md
+++ b/hack/vagrant/README.md
@@ -1,0 +1,29 @@
+# Vagrant Setup to Test the Pouch
+
+This documentation highlights how to use [Vagrant](https://www.vagrantup.com/) to setup node to test pouch.
+
+## Pre-requisites
+
+This was tested on:
+
+- Vagrant 1.9.5
+- VirtualBox 5.1.26
+
+## Getting Started
+
+Clone this repo, change to the `hack/vagrant` directory and let Vagrant do the work.
+
+    $ vagrant up
+    $ vagrant status
+    Current machine states:
+
+    pouch-dev-node            running (virtualbox)
+
+You are now ready to SSH to `pouch-dev-node`, source code of pouch is mounted to /go/src/github.com/alibaba/pouch on `pouch-dev-node`.
+
+    $ vagrant ssh pouch-dev-node
+    vagrant@pouch-dev-node:~$ sudo su
+    root@pouch-dev-node:/home/vagrant# cd /go/src/github.com/alibaba/pouch/
+    root@pouch-dev-node:/go/src/github.com/alibaba/pouch# make && make test
+    ...<snip>...
+
--- a/hack/vagrant/Vagrantfile
+++ b/hack/vagrant/Vagrantfile
@@ -1,0 +1,20 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# refer: https://github.com/hashicorp/vagrant/issues/9442
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
+
+Vagrant.configure("2") do |config|
+  config.vm.boot_timeout=600
+  # ubuntu/xenial64 is from official ubuntu 16.04 LTS (Xenial Xerus) daily build, which is 
+  # based on the long-term supported Linux release series 4.4, refer: https://app.vagrantup.com/ubuntu
+  config.vm.box = "ubuntu/xenial64"
+  config.vm.provision :shell, path: "bootstrap.sh"
+  config.vm.synced_folder "../../", "/go/src/github.com/alibaba/pouch"
+  
+  config.vm.define "pouch-dev-node" do |node|
+    node.vm.hostname = "pouch-dev-node"
+  end
+
+  config.vm.box_check_update = false
+end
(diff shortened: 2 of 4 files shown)


HusterWan:
awsome works,  can your please review this pr @fuweid thanks a lot?

hack/vagrant/README.md
```diff
@@ -0,0 +1,12 @@
+# USAGE
```
fuweid: Could you mind to add more information about Vagrant? For example, the official doc link.

How do you think?

hack/vagrant/bootstrap.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -eo pipefail
+shopt -s nullglob
+
+GOVERSION=1.9.2
```
fuweid: In order to make the shell script readable, we can use function to wrap some common functions. 

Let's make it clear and it's good for maintainers or contributors. Does it make senses?

hack/vagrant/bootstrap.sh
```diff
@@ -0,0 +1,28 @@
...
+SOURCEPATH=/go/src/github.com/alibaba/pouch
+
+# install git/libncurses5-dev/...
+apt-get update
+apt-get install -y git libncurses5-dev libslang2-dev gettext zlib1g-dev libselinux1-dev debhelper lsb-release pkg-config po-debconf autoconf automake autopoint libtool
```
fuweid: This shell command  is too long. We can use the following script to make it readable

```
apt-get install -y git \
    libncurses5-dev \
    .... \
```

hack/vagrant/Vagrantfile
```diff
@@ -0,0 +1,16 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.boot_timeout=600
+  config.vm.box = "bento/ubuntu-16.04"
+  config.vm.provision :shell, path: "bootstrap.sh"
+  config.vm.synced_folder "../../", "/go/src/github.com/alibaba/pouch"
+  
+  config.vm.define "node0" do |node0|
```
fuweid: `node0` looks like random composed name. Can you change it to the `pouch-dev-node` something like that?

hack/vagrant/Vagrantfile
```diff
@@ -0,0 +1,16 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.boot_timeout=600
+  config.vm.box = "bento/ubuntu-16.04"
```
fuweid: Can we use the official image? 
BTW, please offer the kernel version, since the containerd requires the  higher kernel version. If you can offer the information, it maybe help contributor to diagnose.

hack/vagrant/Vagrantfile
```diff
@@ -0,0 +1,22 @@
...
+Vagrant.configure("2") do |config|
+  config.vm.boot_timeout=600
+  # ubuntu/xenial64 is from official ubuntu 16.04 LTS (Xenial Xerus) daily build, which is 
+  # based on the long-term supported Linux release series 4.4, refer: https://app.vagrantup.com/ubuntu
+  config.vm.box = "ubuntu/xenial64"
+  config.vm.provision :shell, path: "bootstrap.sh"
+  config.vm.synced_folder "../../", "/go/src/github.com/alibaba/pouch"
+  
+  config.vm.define "pouch-dev-node" do |node|
+    node.vm.network "private_network", ip: "172.30.30.10"
```
fuweid: Hi, why we need this ip, here? we can use `vagrant ssh` to login, right?

fuweid: Basically, the static IP address is not good. It could cause conflict. If all the envs are the same, that will be fine. However, every dev-env is unique and under different sub-net. 

Therefore, I think we can remove this configure. VirtualBox supports port forwarding so that we can use this to do things which you mentions.

How do you think?

u2takey: You are right, i'll remove this configure and keep it simple. It can be added if then we need a cluster setting like https://github.com/docker/libnetwork/blob/master/Vagrantfile or https://github.com/pires/kubernetes-vagrant-coreos-cluster/blob/master/Vagrantfile.

fuweid: Sure thing!!

hack/vagrant/bootstrap.sh
```diff
@@ -0,0 +1,54 @@
...
+
+    $HOME/.goenv/bin/goenv install $GOVERSION
+    $HOME/.goenv/bin/goenv global $GOVERSION
+}
+
+# install_docker install docker with get.docker.com's shell script
+install_docker(){
+    curl -fsSL get.docker.com -o get-docker.sh
+    sh get-docker.sh
+}
```
fuweid: add the blank line here :)

fuweid:
LGTM. But please use `git commit -s` to add your signed-off. :)

u2takey:
@fuweid thanks, signed.

PR #1264, opened 2018-05-02
bugfix: compability with alidocker when update container diskquota

HusterWan:

alidocker use `Labels["DiskQuota"]` to record container diskquota, so when update diskquota, we also need to update the value of `Labels["DiskQuota"]`


fuweid:
LGTM

PR #1279, opened 2018-05-08
test: make test more robust

Letty5411:

This PR mainly fixed some potential failure when tests run in no-travis-CI  environment.
For example:
* if the `homedir` of pouch daemon is not default value,  we need to get it dynamically.
* If there is legacy volumes, tests like `TestVolumeListOk ` may fail as it checks there are totally and exactly two volumes.
* If users set test images instead of using default ones, test `TestImagesWorks ` may fail.


fuweid:
LGTM

PR #1280, opened 2018-05-08
feature: add timeout handler for execSync in cri part

### LINKED ISSUE #1274: [feature request] add timeout in ExecSync for CRI
### Ⅰ. Issue Description

For now, the `ExecSync` doesn't have timeout functionality. If we don't have this, the long-time exec will consume the pouchd resource, for example, `top/sleep` command. 

Besides this, the `ExecSync` uses the spinning check the status of command. It is not ideally and we should use `pipe` IO to copy and block it. If it is timeout or finish, we will know. Spinning-check isn't good to pouchd.

### Ⅱ. Describe what happened

The long-running process will take pouchd down.

`ExecSync` has the timeout limitation.

1.
2.
3.

@YaoZengzeng 

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ZouRui89:

This pr adds timeout handler for ```execSync``` in cri part. 
I/O pipe mechanism is used to enhance the degree of simplicity, based on which the data type of ```memBuffer``` is changed to allow I/O operations. 
Thus, instead of checking if the execution is finished every 100 millisecond, the error ```io.copy``` returns is used to control the data flow.

fixes #1274

the cri-test has passed.


daemon/mgr/cri.go
```diff
@@ -23,6 +24,7 @@ import (
 	// NOTE: "golang.org/x/net/context" is compatible with standard "context" in golang1.7+.
```
fuweid: Please remove the useless line :)

daemon/mgr/cri.go
```diff
@@ -695,55 +700,71 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
 	execid, err := c.ContainerMgr.CreateExec(ctx, id, createConfig)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create exec for container %q: %v", id, err)
 	}
 
+	reader, writer := io.Pipe()
+
 	var output bytes.Buffer
-	startConfig := &apitypes.ExecStartConfig{}
+	_, err = writer.Write(output.Bytes())
```
fuweid: The pipe doesn't work here.

daemon/mgr/cri.go
```diff
@@ -695,55 +695,85 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
+
 	err = c.ContainerMgr.StartExec(ctx, execid, startConfig, attachConfig)
 	if err != nil {
 		return nil, fmt.Errorf("failed to start exec for container %q: %v", id, err)
 	}
 
-	var execConfig *ContainerExecConfig
-	for {
-		execConfig, err = c.ContainerMgr.GetExecConfig(ctx, execid)
+	var recv bytes.Buffer
```
fuweid: I think we should create goroutine here to make the timeout work.

```
readWaitCh := make(chan error, 1)
go func() {
    defer reader.Close()
    _ , err = io.Copy(&recv, reader)
    readWaitCh <- err
}()

....

select {
case <-ctx.Done():
...
case <-readWaitCh:
...
}
```

daemon/mgr/cri.go
```diff
@@ -695,55 +695,73 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
+	go func() {
+		defer reader.Close()
+		_, err = io.Copy(&recv, reader)
+		if err != nil {
+			logrus.Infof("failed to read data from the pipe")
+		}
+		readWaitCh <- err
+	}()
+
+	select {
```
fuweid: Please add the `ctx.Done()` selector here because the request maybe canceled without waiting the result.

daemon/mgr/cri.go
```diff
@@ -695,55 +695,73 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
+	case <-timeoutCh:
+		return nil, fmt.Errorf("timeout %v exceeded", timeout)
```
fuweid: If this is the timeout, should we stop the exec process? If not, the timeout doesn't make senses here.

daemon/mgr/cri.go
ZouRui89: ```timeoutCh``` is doing the same thing as ctx.Done( ), since I have tried ctx.Done() and some unknown errors will be brought about by that.
And this is the exact way that cri-containerd used to check the ```timeout``` moment, therefore I am following the way it is.

fuweid: Actually, the `ctx.Done()` has been handled by the interceptor. It has the functionality like the `timeoutCh`, but not the same thing. Therefore, could you mind to provide information about what **unknown** error you mentioned?

daemon/mgr/cri.go
ZouRui89: Can you be more specific about the stop execution? I have thought about that but have no idea about how to implement it. :)

daemon/containerio/mem_buffer.go
```diff
@@ -12,7 +11,7 @@ func init() {
 }
 
 type memBuffer struct {
```
YaoZengzeng: Since we have changed the backend from memory buffer to pipe, the name `memBuffer` is no longer properly.

@ZouRui89 Please change to a more appropriate name.

daemon/containerio/options.go
```diff
@@ -102,7 +102,7 @@ func WithStdinHijack() func(*Option) {
 }
 
 // WithMemBuffer specified the memory buffer backend.
-func WithMemBuffer(memBuffer *bytes.Buffer) func(*Option) {
+func WithMemBuffer(memBuffer *io.PipeWriter) func(*Option) {
```
YaoZengzeng: Same as the issue mentioned above.

daemon/mgr/cri.go
```diff
@@ -695,55 +695,82 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
+		if err != nil {
+			logrus.Infof("failed to read data from the pipe")
+		}
+		readWaitCh <- err
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, fmt.Errorf("timeout %v exceeded", timeout)
+	case <-readWaitCh:
```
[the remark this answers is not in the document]
ZouRui89: If the error is not nil, I have used ```logrus.Infof("failed to read data from the pipe")``` to report  the situation.

daemon/mgr/cri.go
```diff
@@ -695,55 +695,73 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
 
-	var execConfig *ContainerExecConfig
-	for {
-		execConfig, err = c.ContainerMgr.GetExecConfig(ctx, execid)
+	readWaitCh := make(chan error, 1)
+	var recv bytes.Buffer
+	go func() {
+		defer reader.Close()
+		_, err = io.Copy(&recv, reader)
+		if err != nil {
```
fuweid: I think we don't need to check the status here. Just put the error into the `readWaitCh` and check it in the receiver's side. As @YaoZengzeng said, we should return the error to the k8s. WDYT

ZouRui89: Right. That sounds more reasonable. :)

PR #1311, opened 2018-05-14
bugfix: make the daemon exit happy

fuweid:
When the daemon server fails, the process will be hang because the
no-one closes the channel. We should exit.

Make the daemon exit happy.

NONE

Basically, we should run the daemon main in new goroutine and wait for it to exit. Based on this, we can receive the signal and clean it up. 

Use available port and it should be success.

```
pouchd --listen tcp://localhost:5000 --debug
```

Use used port and it should be fail

```
python -m SimpleHTTPServer 5000 &
pouchd --listen tcp://localhost:5000 --debug
```

@allencloud , @HusterWan and @rudyfly 

Should we need to clean it up like `sigHandle` when we receive error from `Run()`?

cc @shaloulcy

### DIFF
--- a/main.go
+++ b/main.go
@@ -181,12 +181,26 @@ func runDaemon() error {
 
 	// initialize signal and handle method.
 	var (
-		waitExit = make(chan struct{})
-		signals  = make(chan os.Signal, 1)
+		errCh    = make(chan error, 1)
+		signalCh = make(chan os.Signal, 1)
 	)
-	signal.Notify(signals, syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM, syscall.SIGHUP)
+
+	// new daemon instance, this is core.
+	d := daemon.NewDaemon(cfg)
+	if d == nil {
+		return fmt.Errorf("failed to new daemon")
+	}
+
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -181,37 +181,40 @@ func runDaemon() error {
...
-	return err
+		for _, handle := range sigHandles {
+			if err := handle(); err != nil {
+				logrus.Errorf("failed to handle signal: %v", err)
+			}
+		}
+
+		os.Exit(1)
+	case err := <-errCh:
+		// FIXME: should we do the cleanup like signal handle?
```
HusterWan: I think we also should do cleanup when daemon exited.  
I do not know if this case will happen:  pouchd panic and return error, then we should make the daemon exit gracefully. WDYT ?

rudyfly:
LGTM, and add TODO when server cause error, what to do cleanup.

main.go
fuweid: Agree. But for now, we should hold the scope like @rudyfly said. Refactor the code in the following PR. WDYT?

fuweid: Confirm with @HusterWan , let us add TODO here.

PR #1322, opened 2018-05-15
bugfix: markStopped may failed, will cause container status not right

HusterWan:

`markStoppedAndRelease` do three things now:

  1. change  container status to `stopped`
  2. close container io
  3.  delete container networks

`2` and `3` may failed, it will make container status not changed, so we should ensure container status can be synced as soon as possible


yyb196:
LGTM

PR #1333, opened 2018-05-16
feature: add wait client command for pouch

xiechengsheng:
Add wait client command for pouch

fixes little part of #238

```
$ pouch ps
Name   ID       Status         Created         Image                                            Runtime
foo    f6717e   Up 2 seconds   3 seconds ago   registry.hub.docker.com/library/busybox:latest   runc
$ pouch stop foo
$ pouch ps -a
Name   ID       Status                 Created         Image                                            Runtime
foo    f6717e   Stopped (0) 1 minute   2 minutes ago   registry.hub.docker.com/library/busybox:latest   runc
$ pouch wait foo
0
```
Add more test codes in `test/cli_wait_test.go` and `test/api_container_wait_test.go`


HusterWan:
```
fmt
/usr/local/go/bin/gofmt
test/cli_wait_test.go
Makefile:42: recipe for target 'fmt' failed
make: *** [fmt] Error 1
Exited with code 2
```

CI failed, Please resolve it first, Thanks a lot!!

xiechengsheng:
@HusterWan Oh, I forget to check the code format, thanks for your reply.

cli/wait.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+// runWait is the entry of wait command.
+func (wait *WaitCommand) runWait(args []string) error {
+	ctx := context.Background()
+	apiClient := wait.cli.Client()
+
+	var errs []string
+	for _, name := range args {
```
Letty5411: Could you also do some check for the `args`

test/cli_wait_test.go
```diff
@@ -0,0 +1,106 @@
...
+func (suite *PouchWaitSuite) TearDownTest(c *check.C) {
+}
+
+// TestWaitNonBlockedExitZero is to verify the correctness of waiting a non-blocking container with 0 exit code
+func (suite *PouchWaitSuite) TestWaitNonBlockedExitZero(c *check.C) {
+	name := "TestWaitNonBlockedExitZero"
+	command.PouchRun("run", "-d", "--name", name, busyboxImage, "sh", "-c", "true").Assert(c, icmd.Success)
+	defer DelContainerForceMultyTime(c, name)
+
+	output := command.PouchRun("wait", name).Stdout()
```
Letty5411: I would prefer change this code to:
```
res := command.PouchRun("wait", name)
res. Assert(c, icmd.Success)
output := res.Stdout()
```
In this way, we could get more info if `pouch wait` failed.

xiechengsheng:
Could anyone else help to review this pr? Thanks a lot. 😆

ctrd/container.go
```diff
@@ -500,3 +500,27 @@ func (c *Client) ResizeContainer(ctx context.Context, id string, opts types.Resi
 
 	return pack.task.Resize(ctx, uint32(opts.Height), uint32(opts.Width))
 }
+
+// WaitContainer waits until container's status is stopped.
+func (c *Client) WaitContainer(ctx context.Context, id string, timeout time.Duration) (int64, error) {
```
fuweid: The `waitContainer` API doesn't require the `timeout` param as you mentioned in the swagger.yml. 

I think we can remove the `timeout` here so that it can make `wait` word clear. WDYT?

apis/swagger.yml
```diff
@@ -518,6 +518,29 @@ paths:
...
+        200:
+          description: "The container has exited."
+          schema:
+            type: "object"
+            required: [StatusCode]
+            properties:
+              StatusCode:
+                description: "Exit code of the container"
+                type: "integer"
+                x-nullable: false
```
fuweid: Since the `ProbeContainer` maybe return the unexpected error, I think we should add the error in the response.

test/api_container_wait_test.go
```diff
@@ -0,0 +1,44 @@
...
+	PullImage(c, busyboxImage)
+}
+
+// TestWaitOk tests waiting a stopped container return 200.
+func (suite *APIContainerWaitSuite) TestWaitOk(c *check.C) {
+	cname := "TestWaitOk"
+
+	CreateBusyboxContainerOk(c, cname)
+	StartContainerOk(c, cname)
+	StopContainerOk(c, cname)
```
fuweid: Can we add some cases about running container? I think the common case is using for the `running` container

test/cli_wait_test.go
```diff
@@ -0,0 +1,114 @@
...
+	<-chWait // make sure the goroutine is started
+
+	time.Sleep(100 * time.Millisecond)
+	command.PouchRun("stop", name).Assert(c, icmd.Success)
+
+	select {
+	case status := <-chWait:
+		c.Assert(status, check.Equals, fmt.Sprintf("%s\n", "0"))
+	case <-time.After(2 * time.Second):
+		c.Errorf("timeout waiting for `docker wait` to exit")
```
fuweid: No docker wait, is pouch wait

test/cli_wait_test.go
```diff
@@ -0,0 +1,114 @@
...
+	<-chWait // make sure the goroutine is started
+
+	time.Sleep(100 * time.Millisecond)
+	command.PouchRun("stop", name).Assert(c, icmd.Success)
+
+	select {
+	case status := <-chWait:
+		c.Assert(status, check.Equals, fmt.Sprintf("%s\n", "99"))
+	case <-time.After(2 * time.Second):
+		c.Errorf("timeout waiting for `docker wait` to exit")
```
fuweid: No docker wait, is pouch wait

test/cli_wait_test.go
xiechengsheng: I feel so sorry, this test file is locality referred to docker tests. A little bit embarrassed. :joy:

apis/server/router.go
```diff
@@ -53,6 +53,7 @@ func initRoute(s *Server) http.Handler {
 	s.addRoute(r, http.MethodGet, "/containers/{name:.*}/logs", withCancelHandler(s.logsContainer))
 	s.addRoute(r, http.MethodPost, "/containers/{name:.*}/resize", s.resizeContainer)
 	s.addRoute(r, http.MethodPost, "/containers/{name:.*}/restart", s.restartContainer)
+	s.addRoute(r, http.MethodPost, "/containers/{name:.*}/wait", s.waitContainer)
```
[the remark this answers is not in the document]
xiechengsheng: Yes, you're right~

apis/swagger.yml
fuweid: So, the error means 500?  The error is caused by the container, not the server. The `waitContainer` API doesn't meets any error during the waiting. That is why that I suggest to use the error in the response.

xiechengsheng:
@fuweid Thanks for your review, I have modified many codes according to your suggestions but I don't know whether the latest pr meets with your thought. So could you help to recheck the function `ctrd/container.go WaitContainer()`  please? In my eyes this function codes seems to be a little bit redundant. Do you have a better way to solve this problem? Thanks in advance! 😆

client/interface.go
```diff
@@ -39,6 +39,7 @@ type ContainerAPIClient interface {
 	ContainerTop(ctx context.Context, name string, arguments []string) (types.ContainerProcessList, error)
 	ContainerLogs(ctx context.Context, name string, options types.ContainerLogsOptions) (io.ReadCloser, error)
 	ContainerResize(ctx context.Context, name, height, width string) error
+	ContainerWait(ctx context.Context, name string) (int64, error)
```
fuweid: I think we can use `ContainerWaitOKBody` as return value instead of int64, since the `error` is one of kind error, like `404`, `505` and so on. It can make your code clean. Make senses to you?

ctrd/container.go
```diff
@@ -500,3 +500,27 @@ func (c *Client) ResizeContainer(ctx context.Context, id string, opts types.Resi
...
+func (c *Client) WaitContainer(ctx context.Context, id string) (int64, error) {
+	wrapperCli, err := c.Get(ctx)
+	if err != nil {
+		return -1, fmt.Errorf("failed to get a containerd grpc client: %v", err)
+	}
+
+	ctx = leases.WithLease(ctx, wrapperCli.lease.ID())
+
+	waitExit := func() *Message {
+		return c.ProbeContainer(ctx, id, -1*time.Second)
```
fuweid: I still have concerns about this function `ProbeContainer`, because it consumes the channel receiver's data. If you consumes the data here, some function will loss the data.

fuweid:
Hi @xiechengsheng , thanks for contribution. But I still have some comments. Please take a look. Thanks

client/interface.go
xiechengsheng: OK, I'll try it.

ctrd/container.go
xiechengsheng: @fuweid You're right. But this channel `ch` seems only to be consumed in function `ProbeContainer`. And this function is called in function `DestroyContainer` in order to wait container exits normally. And if you don't mind, I can add additional channel `waitMeg        chan *Message` in struct `containerPack` for `WaitContainer` function to consume specially. I don't have a more elegant way to solve it out...

fuweid: Hi @xiechengsheng , please go ahead and show your code. We can find the way out. 😄

shaloulcy: @xiechengsheng @fuweid ProbeContainer consumes the channel data, and it will push the data back again. So we will not loss the data

cli/wait.go
```diff
@@ -0,0 +1,70 @@
...
+var waitDescription = "Block until one or more containers stop, then print their exit codes. " +
+	"If container state is already stopped, the command will return exit code immediately. " +
+	"On a successful stop, the exit code of the container is returned. "
+
+// WaitCommand is used to implement 'wait' command.
+type WaitCommand struct {
+	baseCommand
+}
+
+// Init initialize wait command.
```
shaloulcy: s/initialize/initializes/

fuweid:
@xiechengsheng good job. Please update the typo and I will merge it

apis/server/container_bridge.go
```diff
@@ -383,3 +383,23 @@ func (s *Server) removeContainers(ctx context.Context, rw http.ResponseWriter, r
...
+
+	if err != nil {
+		return err
+	}
+
+	errMsg := ""
+	if err != nil {
+		errMsg = err.Error()
+	}
+
```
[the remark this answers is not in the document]
fuweid: @xiechengsheng , when you run command and get wrong something from shell, not only you get the error code, but also the error message, like `command not found: xxx`. `errMsg` is used to tell user that `hey, you got something wrong here`. Does it make senses to you? 

In your commit, we can define the struct to wrap error code and error message. The error message is different from the `error` which means the server got something. 

WDYT?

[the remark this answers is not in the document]
xiechengsheng: Got it, thx.

apis/server/container_bridge.go
```diff
@@ -383,3 +383,23 @@ func (s *Server) removeContainers(ctx context.Context, rw http.ResponseWriter, r
...
+	name := mux.Vars(req)["name"]
+
+	status, err := s.ContainerMgr.Wait(ctx, name)
+
+	if err != nil {
+		return err
+	}
+
+	errMsg := ""
+	if err != nil {
```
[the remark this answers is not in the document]
xiechengsheng: @fuweid If `err` is not nil, the HTTP response API should return this `err` directly. If I don't return this `err` and just return `errMsg`, the HTTP response code would be 200, not 404/500.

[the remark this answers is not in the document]
fuweid:
```
if err != nil {
       return err
}

// if the err is not nil, the above statement will return and the following will not execute.
errMsg := ""
if err != nil {
...
}
```

In my options, the `s.ContainerMgr.Wait` should return two vars.

One is 
```
type ErrStatus {
    err error
    errMsg string
}
```

Other one is  error.

Make senses? 😄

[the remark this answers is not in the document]
xiechengsheng: OK, I'll add this.

ctrd/container.go
```diff
@@ -500,3 +500,36 @@ func (c *Client) ResizeContainer(ctx context.Context, id string, opts types.Resi
...
+
+// WaitContainer waits until container's status is stopped.
+func (c *Client) WaitContainer(ctx context.Context, id string) (types.ContainerWaitOKBody, error) {
+	wrapperCli, err := c.Get(ctx)
+	if err != nil {
+		return types.ContainerWaitOKBody{
+			Error:      err.Error(),
+			StatusCode: -1,
+		}, fmt.Errorf("failed to get a containerd grpc client: %v", err)
+	}
```
fuweid: If we got error from `c.Get(ctx)`, we can keep the `ContainerWaitOKBody` simple and return `types.ContainerWaitOKBody{}, fmt.Errorf("failed to get a containerd grpc client: %v", err)`.

In golang, basically we don't consume the data if error != nil

ctrd/container.go
```diff
@@ -500,3 +500,36 @@ func (c *Client) ResizeContainer(ctx context.Context, id string, opts types.Resi
...
+
+	waitExit := func() *Message {
+		return c.ProbeContainer(ctx, id, -1*time.Second)
+	}
+
+	var msg *Message
+	// wait for the task to exit.
+	msg = waitExit()
+
+	if err := msg.RawError(); err != nil && errtypes.IsTimeout(err) {
```
fuweid: If the `errtypes.IsTimeout(err)` is true, we can return `types.ContainerWaitOKBody{}, err` like my last comment. WDYT?

cli/wait.go
```diff
@@ -0,0 +1,72 @@
...
+// runWait is the entry of wait command.
+func (wait *WaitCommand) runWait(args []string) error {
+	ctx := context.Background()
+	apiClient := wait.cli.Client()
+
+	var errs []string
+	for _, name := range args {
+		response, err := apiClient.ContainerWait(ctx, name)
+		if err != nil {
+			errs = append(errs, "container exitcode: "+strconv.FormatInt(response.StatusCode, 10)+
```
fuweid: I think we can just show the error without response, because in golang we don't consume the data if err != nil.

client/container_wait.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/alibaba/pouch/apis/types"
+)
+
+// ContainerWait pauses execution until a container exits.
+// It returns the API status code as response of its readiness.
+func (client *APIClient) ContainerWait(ctx context.Context, name string) (types.ContainerWaitOKBody, error) {
+	resp, err := client.post(ctx, "/containers/"+name+"/wait", nil, nil, nil)
+
+	if err != nil {
+		return types.ContainerWaitOKBody{
```
fuweid: Just return the empty `types.ContainerWaitOKBody` is ok

daemon/mgr/container.go
```diff
@@ -1299,6 +1302,29 @@ func (mgr *ContainerManager) Resize(ctx context.Context, name string, opts types
 	return mgr.Client.ResizeContainer(ctx, c.ID, opts)
 }
 
+// Wait stops processing until the given container is stopped.
+func (mgr *ContainerManager) Wait(ctx context.Context, name string) (types.ContainerWaitOKBody, error) {
+	c, err := mgr.container(name)
+	if err != nil {
+		return types.ContainerWaitOKBody{
```
fuweid: please return the error and empty `types.ContainerWaitOKBody` directly.

daemon/mgr/container.go
```diff
@@ -1299,6 +1302,29 @@ func (mgr *ContainerManager) Resize(ctx context.Context, name string, opts types
...
+			StatusCode: -1,
+		}, err
+	}
+
+	// We should notice that container's meta data shouldn't be locked in wait process, otherwise waiting for
+	// a running container to stop would make other client commands which manage this container are blocked.
+	// If a container status is exited or stopped, return exit code immediately.
+	if c.IsExited() || c.IsStopped() {
+		return types.ContainerWaitOKBody{
+			Error:      "",
```
fuweid: please set the `c.State.Error` for `Error` field.

test/api_container_wait_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	CreateBusyboxContainerOk(c, cname)
+	StartContainerOk(c, cname)
+
+	var (
+		err  error
+		resp *http.Response
+	)
+
+	chWait := make(chan *http.Response)
```
fuweid: tip: if we uses the wait channel for notify, we can use empty struct in golang 😄

test/cli_wait_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// TestWaitBlockedExitZero is to verify the correctness of waiting a blocking container with 0 exit code
+func (suite *PouchWaitSuite) TestWaitBlockedExitZero(c *check.C) {
+	name := "TestWaitBlockedExitZero"
+	command.PouchRun("run", "-d", "--name", name, busyboxImage, "/bin/sh", "-c", "trap 'exit 0' TERM; "+
+		"while true; do usleep 10; done").Assert(c, icmd.Success)
+	defer DelContainerForceMultyTime(c, name)
+
+	chWait := make(chan string)
```
fuweid: tip: please try empty struct for the wait channel

test/api_container_wait_test.go
xiechengsheng: Emmm, I think this channel is not only used for notify but also transferring data to the variable `resp ` in the next. So I think we can't use empty struct here. WDYT?

test/cli_wait_test.go
xiechengsheng: This channel type is `string`, can this be converted to struct{}? And this string channel's data will be transferred to variable `status` in the next.

ctrd/container.go
xiechengsheng: OK, I'll fix all.

ctrd/container.go
fuweid: Solved

ctrd/container.go
fuweid: Solved

test/api_container_wait_test.go
fuweid: Solved

ctrd/container.go
```diff
@@ -500,3 +500,30 @@ func (c *Client) ResizeContainer(ctx context.Context, id string, opts types.Resi
...
+		return types.ContainerWaitOKBody{}, err
+	}
+
+	return types.ContainerWaitOKBody{
+		Error:      "",
```
fuweid: Sorry @xiechengsheng , I miss this one. Please put the `msg.RawError()` to the `Error` field

fuweid:
@xiechengsheng , could you please rebase your branch with latest master branch, since the case has updated by other PR? Thanks

xiechengsheng:
@fuweid Emmmm. I update the branch and repush codes, but it is useless... Seems a testing bug?

fuweid:
Hi @xiechengsheng , could you please to update the `-check.v` to `-check.vv` in `hack/make.sh#271L`? Therefore, we can see which case takes so long.

xiechengsheng:
I'm sorry it's my code's bug......

daemon/mgr/container.go
```diff
@@ -1302,6 +1305,26 @@ func (mgr *ContainerManager) Resize(ctx context.Context, name string, opts types
...
+func (mgr *ContainerManager) Wait(ctx context.Context, name string) (types.ContainerWaitOKBody, error) {
+	c, err := mgr.container(name)
+	if err != nil {
+		return types.ContainerWaitOKBody{}, err
+	}
+
+	// We should notice that container's meta data shouldn't be locked in wait process, otherwise waiting for
+	// a running container to stop would make other client commands which manage this container are blocked.
+	// If a container status is exited or stopped, return exit code immediately.
+	if c.IsExited() || c.IsStopped() {
```
fuweid: How about the `created` container? What is the behaviour?

xiechengsheng: Yeah, the `created` case is very special, this concern is very considerate. So I dive into the code and find that when a new container is created and *`running`*, this map `watch.containers` would record container's pack info. If a container state is created, this code won't go to if statement codeblock and will run to `WaitContainer` function in `ctrd/container.go`, so this `WaitContainer` function will call `ProbeContainer` function, and `ProbeContainer` function will [look up](https://github.com/alibaba/pouch/blob/master/ctrd/container.go#L144) container pack info from the above map `watch.containers`. This function [`watch.notify()`](https://github.com/alibaba/pouch/blob/master/ctrd/watch.go#L140) will look up map `watch.containers`, so when the container is just created, the look up process will [fail](https://github.com/alibaba/pouch/blob/master/ctrd/watch.go#L144) and just return the `Message{}` to channel `ch`, so the `ProbeContainer` function will get `Message{}` channel struct and this struct's member `uint 32 ExitCode` would be initialized with default number `0`. So when waiting a created container, the client will see the code `0`, which is the same as moby.

ctrd/container.go
```diff
@@ -500,3 +500,35 @@ func (c *Client) ResizeContainer(ctx context.Context, id string, opts types.Resi
...
+	var msg *Message
+	// wait for the task to exit.
+	msg = waitExit()
+
+	if err := msg.RawError(); err != nil && errtypes.IsTimeout(err) {
+		return types.ContainerWaitOKBody{}, err
+	}
+
+	errMsg := ""
+	if err := msg.RawError(); err != nil && errtypes.IsTimeout(err) {
```
fuweid: @xiechengsheng , could we use the following statement to make the logic clear? WDYT

```
errMsg := ""
err := msg.RawError()
if err != nil {
       if errtypes.IsTimeout(err) {
		return types.ContainerWaitOKBody{}, err
       }
       errMsg = err.Error()
}
....
```

daemon/mgr/container.go
fuweid: So, can we add the case for this? Thanks

test/api_container_wait_test.go
```diff
@@ -0,0 +1,78 @@
...
+		err  error
+		resp *http.Response
+	)
+
+	chWait := make(chan struct{})
+	go func() {
+		resp, err = request.Post("/containers/" + cname + "/wait")
+		close(chWait)
+	}()
+	time.Sleep(100 * time.Millisecond)
```
fuweid: I think that we have problem for this case. 

`<-chWait` return which means that the channel `chWait` has been closed. In the next statement `select` , the `<-chWait` always return.

```
package main

import (
        "fmt"
        "time"
)

func main() {
        ch := make(chan struct{})

        go func() {
                time.Sleep(2 * time.Second)
                close(ch)
        }()
        fmt.Println(<-ch)
        fmt.Println(<-ch)
}
```

 So I think we still need to update the case like, WDYT?

```
chWait := make(chan struct{})
go func() {
        chWait <- struct{}{}
        resp, err = request.Post("/containers/" + cname + "/wait")
        close(chWait)
}()
<-chWait // wait for the goroutine activate
time.Sleep(100 * time.Millisecond)
stopContainerOk(c, cname)

select {
...
}
```

test/api_container_wait_test.go
```diff
@@ -0,0 +1,78 @@
...
+	resp, err := request.Post("/containers/" + cname + "/wait")
+	c.Assert(err, check.IsNil)
+	CheckRespStatus(c, resp, 200)
+
+	DelContainerForceOk(c, cname)
+}
+
+// TestRunningContainer tests waiting a running container to stop, then returns 200.
+func (suite *APIContainerWaitSuite) TestRunningContainer(c *check.C) {
+	cname := "TestRunningContainer"
```
fuweid: Please use the `TestWaitForRunningContainer` so that the case doesn't impact other case.

ctrd/container.go
HusterWan: I think it's ok, only if the code is clear in logic

test/api_container_wait_test.go
HusterWan: Agree! @xiechengsheng please change the function name, because `TestRunningContainer` may has confusion, thanks a lot !

test/api_container_wait_test.go
HusterWan: since we have call `<-chWait` before, why we wait `chWait` in `select` again ?

xiechengsheng:
@fuweid OK, I'll fix all ASAP.

ctrd/container.go
fuweid: emmm... could we remove this duplicate line, since you have done it in the following statement?

daemon/mgr/container.go
```diff
@@ -1302,6 +1305,33 @@ func (mgr *ContainerManager) Resize(ctx context.Context, name string, opts types
...
+		return types.ContainerWaitOKBody{
+			Error:      c.State.Error,
+			StatusCode: c.ExitCode(),
+		}, nil
+	}
+	// If a container status is created, return 0 as status code.
+	if c.IsCreated() {
+		return types.ContainerWaitOKBody{
+			Error:      c.State.Error,
```
fuweid: I think we can just left the Error field empty. WDYT?

xiechengsheng: OK, I'll fix all.

fuweid:
LGTM. I will merge it after CI is green. 

Thanks @xiechengsheng ! Great job.

PR #1354, opened 2018-05-18
bugfix: remove pids-limit initial value

Ace-Tang:
pids-limit should have zero as initial value, or runtime
will fail to set if kernel not support pid-limit, that will
fail to start container.

### DIFF
--- a/cli/common_flags.go
+++ b/cli/common_flags.go
@@ -84,7 +84,7 @@ func addCommonFlags(flagSet *pflag.FlagSet) *container {
 
 	flagSet.StringVarP(&c.workdir, "workdir", "w", "", "Set the working directory in a container")
 	flagSet.Var(&c.ulimit, "ulimit", "Set container ulimit")
-	flagSet.Int64Var(&c.pidsLimit, "pids-limit", -1, "Set container pids limit, -1 for unlimited")
+	flagSet.Int64Var(&c.pidsLimit, "pids-limit", 0, "Set container pids limit")
 
 	flagSet.BoolVar(&c.rich, "rich", false, "Start container in rich container mode. (default false)")
 	flagSet.StringVar(&c.richMode, "rich-mode", "", "Choose one rich container mode. dumb-init(default), systemd, sbin-init")


cli/common_flags.go
```diff
@@ -84,7 +84,7 @@ func addCommonFlags(flagSet *pflag.FlagSet) *container {
 
 	flagSet.StringVarP(&c.workdir, "workdir", "w", "", "Set the working directory in a container")
 	flagSet.Var(&c.ulimit, "ulimit", "Set container ulimit")
-	flagSet.Int64Var(&c.pidsLimit, "pids-limit", -1, "Set container pids limit, -1 for unlimited")
+	flagSet.Int64Var(&c.pidsLimit, "pids-limit", 0, "Set container pids limit, -1 for unlimited")
```
fuweid: I check the [document](https://www.kernel.org/doc/Documentation/cgroup-v1/pids.txt) and I think the unlimited setting should be `max`. 

It's correct? 

I check this in my env:

```
➜  ~ cat /sys/fs/cgroup/pids/user.slice/pids.max
max
```

fuweid: And please update the comment for the `pidsLimit`.

Ace-Tang: you are right, comment maybe confused, -1 here just means in runc, we would not set value for pid.max , it will jump the set.  Maybe we can add more comment, but -1 here is really for unlimit set, @fuweid 
```
if cgroup.Resources.PidsLimit != 0 {
 31         // "max" is the fallback value.
 32         limit := "max"
 33 
 34         if cgroup.Resources.PidsLimit > 0 {
 35             limit = strconv.FormatInt(cgroup.Resources.PidsLimit, 10)
 36         }
 37 
 38         if err := writeFile(path, "pids.max", limit); err != nil {
 39             return err
 40         }
 41     }
```

Ace-Tang: I think I get something wrong.

Ace-Tang: Oh, -1 means will still set max for pids.max, but if pids cgroup is not mount, it will go error, let me re-write the comment.

fuweid: Nice

fuweid:
LGTM

fuweid:
BTW, I think we should add some tests for the cgroup settings in the next step.

Ace-Tang:
Yes, cgroup file existing should be check first

PR #1361, opened 2018-05-21
bugfix: fix cgroup-parent can not set from config

Ace-Tang:
remove initial value in cgroup-parent flag, or this flag
can never be set from config file when create container
with pouch cli.

```
func (mgr *ContainerManager) createContainerdContainer(ctx context.Context, c *Container) error {
    // CgroupParent from HostConfig will be first priority to use,
    // then will be value from mgr.Config.CgroupParent
    if c.HostConfig.CgroupParent == "" { 
        c.HostConfig.CgroupParent = mgr.Config.CgroupParent
    }    
```
If pouch cli give cgroup-parent a default value, the `c.HostConfig.CgroupParent ` will  never be empty, and it can never use `CgroupParent ` set from daemon config file.

The default cgroup parent we already set in 
```
    // same with containerd use. or make it a variable
    cgroupsParent := "default"
    if c.HostConfig.CgroupParent != "" {
        cgroupsParent = c.HostConfig.CgroupParent
    }   
```

### DIFF
--- a/cli/common_flags.go
+++ b/cli/common_flags.go
@@ -91,7 +91,7 @@ func addCommonFlags(flagSet *pflag.FlagSet) *container {
 	flagSet.StringVar(&c.initScript, "initscript", "", "Initial script executed in container")
 
 	// cgroup
-	flagSet.StringVarP(&c.cgroupParent, "cgroup-parent", "", "default", "Optional parent cgroup for the container")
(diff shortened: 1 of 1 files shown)


HusterWan:
Reasonable !!!, Thanks a lot @Ace-Tang 

LGTM

PR #1375, opened 2018-05-22
feature: add cgroup resources check

Ace-Tang:
check cgroup resource valid in container create,
discard un-support cgroup set related flag.

Pouch run get lots of flags that whether these flags can be set depend on the 
corresponding cgroup file is exist. Add check for the cgroup file is needed.

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -360,6 +360,12 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 		container.Snapshotter.Data["UpperDir"] = upperDir
 	}
 
+	// validate container Config
+	warnings, err := validateConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
 	container.Lock()
 	defer container.Unlock()
 
@@ -374,8 +380,9 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 	mgr.cache.Put(id, container)
 
 	return &types.ContainerCreateResp{
-		ID:   id,
-		Name: name,
+		ID:       id,
+		Name:     name,
+		Warnings: warnings,
 	}, nil
 }
 
--- a/pkg/system/cgroup.go
+++ b/pkg/system/cgroup.go
@@ -1,0 +1,143 @@
+package system
+
+import (
+	"bufio"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+// MemoryCgroupInfo defines memory cgroup information on current machine
+type MemoryCgroupInfo struct {
+	MemoryLimit      bool
+	MemorySwap       bool
+	MemorySwappiness bool
+	OOMKillDisable   bool
+}
+
+// CPUCgroupInfo defines cpu cgroup information on current machine
+type CPUCgroupInfo struct {
+	CpusetCpus bool
+	CpusetMems bool
+	CPUShares  bool
+	CPUPeriod  bool
+	CPUQuota   bool
+}
+
+// BlkioCgroupInfo defines blkio cgroup information on current machine
+type BlkioCgroupInfo struct {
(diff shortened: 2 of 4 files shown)


HusterWan:
@Ace-Tang please create PR base on branch 0.5.x, thanks a lot !!! 💯

pkg/system/cgroup.go
```diff
@@ -0,0 +1,143 @@
...
+	return &CgroupInfo{
+		Memory: getMemoryCgroupInfo(cgroupRootPath),
+		CPU:    getCPUCgroupInfo(cgroupRootPath),
+		Blkio:  getBlkioCgroupInfo(cgroupRootPath),
+		Pids:   getPidsCgroupInfo(cgroupRootPath),
+	}
+}
+
+func getMemoryCgroupInfo(root string) *MemoryCgroupInfo {
+	path := root + "/" + "memory"
```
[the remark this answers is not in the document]
Ace-Tang: Or `root + "/memory"` more single?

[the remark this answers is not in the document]
allencloud: Actually I do not think `root + "/memory"` is a good way. String append via `+` is not very readable and brings some difficulty to code review. While I think for path, we can use `path.Join` like @rudyfly suggested. For others, I suggested that we should use format to make it like: `fmt.Sprintf("you told me %s", msg)`.

So I think we had better update this.

daemon/mgr/container.go
```diff
@@ -360,6 +360,12 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 		container.Snapshotter.Data["UpperDir"] = upperDir
 	}
 
+	// validate container Config
+	warnings, err := validateConfig(config)
```
allencloud: I think this part validation could be located in the API bridge layer. WDYT? @Ace-Tang @fuweid

fuweid: @allencloud the config has changed by `ContainerManager).Create`, like 

```
if config.HostConfig.EnableLxcfs && lxcfs.IsLxcfsEnabled {
 		config.HostConfig.Binds = append(config.HostConfig.Binds, lxcfs.LxcfsParentDir+":/var/lib/lxc:shared")
 		sourceDir := lxcfs.LxcfsHomeDir + "/proc/"
 		destDir := "/proc/"
 		for _, procFile := range lxcfs.LxcfsProcFiles {
 			bind := fmt.Sprintf("%s%s:%s%s", sourceDir, procFile, destDir, procFile)
 			config.HostConfig.Binds = append(config.HostConfig.Binds, bind)
 		}
 	}
```

I don't think we can do validation in the API bridge level.

Ace-Tang: @fuweid ,do you mean config is merged with something after `ContainerManager).Create`?

Ace-Tang:
@allencloud @rudyfly @fuweid , updated as comment, but except `validate location`, as @fuweid said, config will changed in `Create` funtion.

daemon/mgr/container.go
fuweid: Yes @Ace-Tang . If we can move the merge function into bridge, the validation can be done in bridge, too.

Ace-Tang:
CRI ci test fail

daemon/mgr/container.go
```diff
@@ -374,8 +380,9 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 	mgr.cache.Put(id, container)
 
 	return &types.ContainerCreateResp{
-		ID:   id,
-		Name: name,
+		ID:       id,
+		Name:     name,
+		Warnings: warnings,
```
allencloud: If we already add this part in the daemon's API, is there any possibility that we should support this field in command line of `pouch create` command? @Ace-Tang

allencloud:
LGTM

PR #1378, opened 2018-05-22
feature: init pouch tag functionality

fuweid:
Since the containerd provides the ImageService to modify the boltdb
data, we can add tag reference into the containerd. Just in case that
the client removes the source image.

Use containerd's ImageService to implement `pouch tag` functionality. Allow user to create tag reference to existing image.

NONE.

The containerd stores the image name as key in the `boltdb`. `pouch run/image inspect/rmi` will use the primary reference as name to handle image information.

If we don't use the `boltdb` provided by containerd.ImageService, we will loss the tag reference when we remove the source image.

Therefore, when user adds tag reference to the existing image, I will do the following things:
* get the image from containerd
* use the tag reference as primary name and update boltdb by containerd.ImageService

The update action will add the reference count for the image content. That is the key to implement this functionality.

First one is the help:

```
➜  pouch git:(feature_image_tag) pouch tag -h
tag command is to add tag reference for the existing image.

Usage:
  pouch tag SOURCE_IMAGE[:TAG] TARGET_IMAGE[:TAG]

Examples:
$ pouch tag registry.hub.docker.com/library/busybox:1.28 busybox:latest

Flags:
  -h, --help   help for tag

Global Flags:
  -D, --debug              Switch client log level to DEBUG mode
  -H, --host string        Specify connecting address of Pouch CLI (default "unix:///var/run/pouchd.sock")
      --tlscacert string   Specify CA file of TLS
      --tlscert string     Specify cert file of TLS
      --tlskey string      Specify key file of TLS
      --tlsverify          Use TLS and verify remote
```

Use Tag:

```
➜  pouch git:(feature_image_tag) pouch images
IMAGE ID       IMAGE NAME                                     SIZE
8ac48589692a   registry.hub.docker.com/library/busybox:1.28   710.83 KB
➜  pouch git:(feature_image_tag) pouch tag busybox:1.28 busybox:test
➜  pouch git:(feature_image_tag) pouch images --digest
IMAGE ID       IMAGE NAME                                     DIGEST                                                                    SIZE
8ac48589692a   registry.hub.docker.com/library/busybox:1.28   sha256:58ac43b2cc92c687a32c8be6278e50a063579655fe3090125dcb2af0ff9e1a64   710.83 KB
8ac48589692a   registry.hub.docker.com/library/busybox:test   sha256:58ac43b2cc92c687a32c8be6278e50a063579655fe3090125dcb2af0ff9e1a64   710.83 KB
```

We can use the tag reference to run container like:

```
➜  pouch git:(feature_image_tag) pouch run busybox:test ls
bin
dev
etc
home
proc
root
run
sys
tmp
usr
var
```

But we don't allow to use `sha256` as name, because it will confuse with searching by imageID.

```
➜  pouch git:(feature_image_tag) pouch tag busybox:1.28 sha256:oops
Error: {"message":"refusing to create an reference using digest algorithm as name: invalid param"}
```

However, it doesn't act like `docker` right now. We don't allow user to override the existing primary reference, like: 

```
➜  pouch git:(feature_image_tag) pouch images
IMAGE ID       IMAGE NAME                                     SIZE
e02e811dd08f   registry.hub.docker.com/library/busybox:1.25   654.44 KB
8ac48589692a   registry.hub.docker.com/library/busybox:1.28   710.83 KB
8ac48589692a   registry.hub.docker.com/library/busybox:test   710.83 KB
➜  pouch git:(feature_image_tag) pouch tag e02e811dd08f busybox:test # even if the busybox:test is created by tag
Error: {"message":"the tag reference (registry.hub.docker.com/library/busybox:test) has been used as reference: invalid param"}
```

### DIFF
--- a/apis/server/image_bridge.go
+++ b/apis/server/image_bridge.go
@@ -117,3 +117,20 @@ func (s *Server) removeImage(ctx context.Context, rw http.ResponseWriter, req *h
 	rw.WriteHeader(http.StatusNoContent)
 	return nil
 }
+
+// postImageTag adds tag for the existing image.
+func (s *Server) postImageTag(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
+	name := mux.Vars(req)["name"]
+
+	targetRef := req.FormValue("repo")
+	if tag := req.FormValue("tag"); tag != "" {
+		targetRef = fmt.Sprintf("%s:%s", targetRef, tag)
+	}
+
+	if err := s.ImageMgr.AddTag(ctx, name, targetRef); err != nil {
+		return err
+	}
+
+	rw.WriteHeader(http.StatusCreated)
+	return nil
+}
--- a/apis/server/router.go
+++ b/apis/server/router.go
@@ -60,6 +60,7 @@ func initRoute(s *Server) http.Handler {
 	s.addRoute(r, http.MethodGet, "/images/json", s.listImages)
 	s.addRoute(r, http.MethodDelete, "/images/{name:.*}", s.removeImage)
 	s.addRoute(r, http.MethodGet, "/images/{name:.*}/json", s.getImage)
+	s.addRoute(r, http.MethodPost, "/images/{name:.*}/tag", s.postImageTag)
 
 	// volume
 	s.addRoute(r, http.MethodGet, "/volumes", s.listVolume)
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -265,6 +265,34 @@ paths:
         500:
           $ref: "#/responses/500ErrorResponse"
 
+  /images/{imageid}/tag:
+    post:
+      summary: "Tag an image"
+      description: "Add tag reference to the existing image"
+      parameters:
+        - $ref: "#/parameters/imageid"
+        - name: "repo"
+          in: "query"
+          description: "The repository to tag in. For example, `someuser/someimage`."
+          type: "string"
+        - name: "tag"
+          in: "query"
+          description: "The name of the new tag."
+          type: "string"
+      responses:
+        201:
+          description: "No error"
+        400:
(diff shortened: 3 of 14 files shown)


apis/server/image_bridge.go
```diff
@@ -117,3 +117,20 @@ func (s *Server) removeImage(ctx context.Context, rw http.ResponseWriter, req *h
 	rw.WriteHeader(http.StatusNoContent)
 	return nil
 }
+
+// postImageTag adds tag for the existing image.
+func (s *Server) postImageTag(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
+	name := mux.Vars(req)["name"]
+
+	targetRef := req.FormValue("repo")
+	if tag := req.FormValue("tag"); tag != "" {
```
HusterWan: if tag is "", do we need set default value `latest`?

apis/server/router.go
```diff
@@ -60,6 +60,7 @@ func initRoute(s *Server) http.Handler {
 	s.addRoute(r, http.MethodGet, "/images/json", s.listImages)
 	s.addRoute(r, http.MethodDelete, "/images/{name:.*}", s.removeImage)
 	s.addRoute(r, http.MethodGet, "/images/{name:.*}/json", s.getImage)
+	s.addRoute(r, http.MethodPost, "/images/{name:.*}/tag", s.postImageTag)
```
HusterWan: need add swagger info

apis/server/image_bridge.go
fuweid: I have handled it in the deamon so that we can keep the bridge simple.

apis/server/router.go
fuweid: Good catch.

HusterWan:
LGTM

PR #1386, opened 2018-05-23
bugfix: make the PullImage test util work

fuweid:

The post method of pullImage will hijack the connection so that the HTTP header always contains 200. It means that we cannot just check the HTTP status. We should read all the data bytes from server.

NONE

Read the bytes from server and check the error in the jsonstream.

Remove the `command.PouchRun("pull", busyboxImage)` in the beginning of test so that we can make sure the `PullImage` works well or not.

It is very important to make CI happy

### DIFF
--- a/test/main_test.go
+++ b/test/main_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/alibaba/pouch/client"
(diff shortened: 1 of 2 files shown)


Letty5411:
LGTM

PR #1400, opened 2018-05-24
enhance: clean the local-persist after integration tests

shaloulcy:

After integration tests, we will clean the local-persist volume plugin

#1396

### DIFF
--- a/hack/make.sh
+++ b/hack/make.sh
@@ -117,6 +117,21 @@ function install_local_persist
 	mv /tmp/local-persist /usr/local/bin/
 }
 
+# clean the local-persist
+function clean_local_persist
+{
+	echo "Try cleaning local-persist"
+	pid=$(pgrep local-persist)
+
+	if [[ $pid ]]; then
+		echo "Try killing local-persist process"
+		kill -9 $pid
+	fi
+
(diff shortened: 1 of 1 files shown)


hack/make.sh
```diff
@@ -117,6 +117,21 @@ function install_local_persist
 	mv /tmp/local-persist /usr/local/bin/
 }
 
+# clean the local-persist
+function clean_local_persist
+{
+	echo "Try cleaning local-persist"
+	pid=`ps axu|grep local-persist |grep -v grep|awk '{print $2}'`
```
Letty5411: How about just pgrep ?

hack/make.sh
```diff
@@ -117,6 +117,21 @@ function install_local_persist
...
+
+	if [ -n "$pid" ]; then
```
Letty5411: use [[ ]] instead of [ ]

hack/make.sh
shaloulcy: done it

hack/make.sh
shaloulcy: done it

rudyfly:
LGTM

PR #1406, opened 2018-05-24
bugfix: use busybox:1.25 instead of busybox:1.28 in cli_tag_test

fuweid:


HusterWan:
LGTM

PR #1447, opened 2018-05-31
fix: correct shell script format via shellcheck reports

zhuangqh:

This PR help pouch to pass the ShellCheck (https://github.com/koalaman/shellcheck). There is a short report.

SC2086 Double quote to prevent globbing and word splitting **94** fail(s)
SC2046 Quote this to prevent word splitting **8** fail(s)
SC2034 POUCH_SOCK appears unused. Verify it or export it **7** fail(s)
SC2039 In POSIX sh, 'pushd' is not supported **6** fail(s)
SC2148 Tips depend on target shell and yours is unknown. Add a shebang **5** fail(s)
SC2112 'function' keyword is non-standard. Delete it **4** fail(s)
SC2016 Expressions don't expand in single quotes, use double quotes for that **3** fail(s)
SC2006 Use $(..) instead of legacy `..` **2** fail(s)
SC2129 Consider using { cmd1; cmd2; } >> file instead of individual redirects **1** fail(s)
SC2071 < is for string comparisons. Use -lt instead **1** fail(s)
SC2068 Double quote array expansions, otherwise they're like $* and break on spaces **1** fail(s)
SC2064 Use single quotes, otherwise this expands now rather than when signalled **1** fail(s)

You can find the detailed rules [here](https://github.com/koalaman/shellcheck/wiki)

ShellCheck don't report error. 
```bash
find ./ -name "*.sh" | grep -v vendor | grep -v extra | xargs shellcheck
```
Shell script still works fine.

In rule SC2086
> When quoting composite arguments, make sure to exclude globs and brace expansions, which lose their special meaning in double quotes: "$HOME/$dir/src/\*.c" will not expand, but "$HOME/$dir/src"/\*.c will.

### DIFF
--- a/hack/make.sh
+++ b/hack/make.sh
@@ -6,7 +6,7 @@ set -ex
 #
 TMP=$(mktemp -d /tmp/pouch.XXXXXX)
 DIR="$( cd "$( dirname "$0" )/.." && pwd )"
-cd $DIR/
+cd "$DIR/"
 
 CONTAINERD_VERSION=
 RUNC_VERSION=
@@ -19,8 +19,8 @@ IMAGE="registry.hub.docker.com/letty/pouchci:latest"
 if [[ $SOURCEDIR != $DIR ]];then
 	[ -d $SOURCEDIR ] && rm -rf $SOURCEDIR
 	POUCHTOPDIR=$(dirname $SOURCEDIR)
-	[ ! -d $POUCHTOPDIR ] && mkdir -p $POUCHTOPDIR
-	ln -sf $DIR/ $SOURCEDIR
+	[ ! -d "$POUCHTOPDIR" ] && mkdir -p "$POUCHTOPDIR"
+	ln -sf "$DIR/" $SOURCEDIR
 fi
 
 #
@@ -71,9 +71,9 @@ function install_containerd
 	else
 		echo "Download and install containerd."
 		wget --quiet \
-			https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P $TMP
-		tar xf $TMP/containerd-1.0.3.linux-amd64.tar.gz -C $TMP &&
-			cp -f $TMP/bin/* /usr/local/bin/
+			https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P "$TMP"
+		tar xf "$TMP/containerd-1.0.3.linux-amd64.tar.gz" -C "$TMP" &&
+			cp -f "$TMP"/bin/* /usr/local/bin/
 	fi;
 }
 
@@ -103,7 +103,7 @@ function install_lxcfs
 			apt-get install -y lxcfs
 		fi
 	else
-		sh -x $DIR/hack/install_lxcfs_on_centos.sh
+		sh -x "$DIR/hack/install_lxcfs_on_centos.sh"
 	fi
 }
 
@@ -125,7 +125,7 @@ function clean_local_persist
 
 	if [[ $pid ]]; then
 		echo "Try killing local-persist process"
-		kill -9 $pid
+		kill -9 "$pid"
 	fi
 
 	echo "Try removing local-persist.sock"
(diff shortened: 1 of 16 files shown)


fuweid:
Nice!

zhuangqh:
related to #1437

hack/package/rpm/build.sh
```diff
@@ -79,13 +78,13 @@ function build_rpm ()
...
-          -a $ARCHITECTURE \
-          -p $PACKAGEDIR \
+          -v "$VERSION" \
+          --iteration "$ITERATION" \
+          -a "$ARCHITECTURE" \
+          -p "$PACKAGEDIR" \
           --description 'Pouch is an open-source project created by Alibaba Group to promote the container technology movement.
 
-    Pouchs vision is to advance container ecosystem and promote container standards OCI, so that container technologies become the foundation for application development in the Cloud era.
+    Pouch'"'"'s vision is to advance container ecosystem and promote container standards OCI, so that container technologies become the foundation for application development in the Cloud era.
```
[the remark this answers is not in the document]
zhuangqh: This change aims to display single quote correctly. @allencloud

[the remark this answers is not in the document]
allencloud: Oh, I see. This is not just doc, but a string in shell script. My fault.

hack/make.sh
```diff
@@ -71,9 +71,9 @@ function install_containerd
 	else
 		echo "Download and install containerd."
 		wget --quiet \
-			https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P $TMP
-		tar xf $TMP/containerd-1.0.3.linux-amd64.tar.gz -C $TMP &&
-			cp -f $TMP/bin/* /usr/local/bin/
+			https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P "$TMP"
+		tar xf "$TMP/containerd-1.0.3.linux-amd64.tar.gz" -C "$TMP" &&
```
allencloud: I think in the future we need to make this version configurable at the head of this file rather than hard coding here.

hack/package/rpm/build.sh
```diff
@@ -48,19 +47,19 @@ function build_pouch()
 {
     # install containerd
     echo "Downloading containerd."
-    wget --quiet https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P $TMP
-    tar xf $TMP/containerd-1.0.3.linux-amd64.tar.gz -C $TMP && cp -f $TMP/bin/* $BINDIR/
+    wget --quiet https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P "$TMP"
```
allencloud: Also this place, there is hard coding

hack/package/rpm/build.sh
```diff
@@ -48,19 +47,19 @@ function build_pouch()
...
+    tar xf "$TMP/containerd-1.0.3.linux-amd64.tar.gz -C $TMP && cp -f $TMP/bin/* $BINDIR/"
 
     # install runc
     echo "Downloading runc."
-    wget --quiet https://github.com/alibaba/runc/releases/download/v1.0.0-rc4-1/runc.amd64 -P $BINDIR/
-    chmod +x $BINDIR/runc.amd64
-    mv $BINDIR/runc.amd64 $BINDIR/runc
+    wget --quiet https://github.com/alibaba/runc/releases/download/v1.0.0-rc4-1/runc.amd64 -P "$BINDIR/"
```
allencloud: also here, hard coding

allencloud:
LGTM since I take a quick review of this. However, I still wish to invite @ZouRui89 to take a thorough review of this. Thanks a lot.

hack/package/rpm/build.sh
```diff
@@ -48,19 +47,19 @@ function build_pouch()
 {
     # install containerd
     echo "Downloading containerd."
-    wget --quiet https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P $TMP
-    tar xf $TMP/containerd-1.0.3.linux-amd64.tar.gz -C $TMP && cp -f $TMP/bin/* $BINDIR/
+    wget --quiet https://github.com/containerd/containerd/releases/download/v1.0.3/containerd-1.0.3.linux-amd64.tar.gz -P "$TMP"
+    tar xf "$TMP/containerd-1.0.3.linux-amd64.tar.gz -C $TMP && cp -f $TMP/bin/* $BINDIR/"
```
zhuangqh: There should double quote one by one, or won't work.😅

zhuangqh: I will fix this in another PR, together with the problem of hardcode.

fuweid: @zhuangqh cool!

PR #1458, opened 2018-06-04
stdio: add stdin fifo

Ace-Tang:
we should add stdin fifo for compatible of docker 1.12


YaoZengzeng:
@Ace-Tang
 From my point of view, we should open the stdin only when we ask it todo like: `pouch -i run busybox`.

For CRI part, some of the node-e2e test cases will hang if open the stdin by default which is not the same as expected.

So could we make the opt.stdin be true out of the Container Manager, so it will not influence the CRI part? 😄

Ace-Tang:
@YaoZengzeng， Absolutely，we should not influence CRI part.

Ace-Tang:
@YaoZengzeng , @HusterWan , PTAL

Issue #1460, opened 2018-06-04
bug: pouch not pull containerd when containerd get killed


Ace-Tang:
### Ⅰ. Issue Description

pouch not pull containerd when containerd get killed

### Ⅱ. Describe what happened

pouch pull containerd

### Ⅲ. Describe what you expected to happen

pouch not pull containerd

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.  start pouch , `pouchd`
2.  kill containerd, `kill containerd`
3.  check containerd is be pulled by pouchd , `ps -ef | grep pouch`

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Ace-Tang:
I am working on it.

Ace-Tang:
Design of keep pull containerd when it get accident.
At first I think about using containerd event subscribe function to instead of task wait to wait task finish signal,  but the result did not show it will be a better way,  the only benefit take by event maybe is we do not care about task status when containerd is killed unexpected. Because in task wait way, when  containerd is killed unexpected, task status channel will get a channel broken message `rpc error: code = Internal desc = transport is closing`.

Since event design not take more better, we still use task wait way. In design, when containerd is killed unexpected, `handleConnectContainerd` will restart containerd and restore task, since task status channel has broken, so we make a new channel for task, and do re-wait for task exit.

Ace-Tang:
By the way, I do not get what is use for function `isContainerdDead` and `setContainerdDead` , when containerd killed unexpected,  task wait channel receives broken message's is not sure, it can be before or later when containerd is dead completely，so there is no good timing to `setContainerdDead`. But to choose a zero timestamps to decided whether a exit message is useful.
about `setContainerdDead`,  WDYT @HusterWan .

PR #1475, opened 2018-06-06
bugfix: fix pouch not restore containerd when containerd killed unexpectly

Ace-Tang:
fix pouch not restore containerd when it get accident，refactor
the way in wait task exit, detail is in
https://github.com/alibaba/pouch/issues/1460

Fixes: #1460

### DIFF
--- a/ctrd/client.go
+++ b/ctrd/client.go
@@ -44,7 +44,10 @@ type Client struct {
 	pool      []scheduler.Factory
 	scheduler scheduler.Scheduler
 
-	hooks []func(string, *Message) error
+	hooks   []func(string, *Message) error
+	options clientOpts
+	cmdDone chan struct{}
+	isStart bool
 }
 
 // NewClient connect to containerd.
@@ -74,11 +77,13 @@ func NewClient(homeDir string, opts ...ClientOpt) (APIClient, error) {
 		oomScoreAdjust: copts.oomScoreAdjust,
 		debugLog:       copts.debugLog,
 		rpcAddr:        copts.rpcAddr,
+		options:        copts,
+		cmdDone:        make(chan struct{}, 1),
 	}
 
 	// start new containerd instance.
 	if copts.startDaemon {
-		if err := client.runContainerdDaemon(homeDir, copts); err != nil {
+		if err := client.runContainerdDaemon(); err != nil {
 			return nil, err
 		}
 	}
@@ -98,6 +103,10 @@ func NewClient(homeDir string, opts ...ClientOpt) (APIClient, error) {
 		return nil, fmt.Errorf("failed to create clients pool scheduler")
 	}
 	client.scheduler = scheduler
+
+	client.isStart = true
+	// handle containerd failure
+	go client.handleConnectContainerd()
 
 	return client, nil
 }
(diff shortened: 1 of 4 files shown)


Ace-Tang:
Detail design logic is locate at https://github.com/alibaba/pouch/issues/1460

rudyfly:
ping @fuweid @HusterWan Can you review this pr?

ctrd/client.go
```diff
@@ -328,3 +346,30 @@ func (c *Client) Cleanup() error {
 
 	return nil
 }
+
+// handleConnectContainerd restore containerd
+func (c *Client) handleConnectContainerd() {
+
+	for {
+		logrus.Info("handle restore containerd")
```
HusterWan: > logrus.Info("handle restore containerd")

> logrus.Infof("ctrd: restore containerd")

should we only keep one log?

ctrd/client.go
```diff
@@ -328,3 +346,30 @@ func (c *Client) Cleanup() error {
 
 	return nil
 }
+
+// handleConnectContainerd restore containerd
+func (c *Client) handleConnectContainerd() {
```
HusterWan: If containerd process restart, we should also recreate new ctrd client,  WDYT?

HusterWan:
@Ace-Tang Thanks a lot for your work to enhance the management of containerd process instance.

There is indeed a shortcut that when containerd process existed unexpected, we will not pull it up again before, so i think this pr is wonderful !!

there has one thing we should discuss, the old grpc client's stream client has 100 limitation, so we use a grpc clients pool to break the limitation.

so if we restart containerd, we also should notice the stream client counter.

ctrd/client.go
Ace-Tang: Sure.

ctrd/client.go
Ace-Tang: Maybe client pool need more optimization.

test/cli_exec_test.go
```diff
@@ -143,3 +147,24 @@ func (suite *PouchExecSuite) TestExecExitCode(c *check.C) {
 	command.PouchRun("exec", name, "sh", "-c", "exit 101").Assert(c, icmd.Expected{ExitCode: 101})
 	command.PouchRun("exec", name, "sh", "-c", "exit 0").Assert(c, icmd.Success)
 }
+
+// TestExecWithContainerdKilled test containerd get unexpected killed, and will restore
+func (suite *PouchExecSuite) TestExecWithContainerdKilled(c *check.C) {
+	name := "TestExecWithContainerdKilled"
+	res := command.PouchRun("run", "-d", "--name", name, busyboxImage, "top")
+	res.Assert(c, icmd.Success)
```
Letty5411: add a `defer DelContainerForceMultyTime(c, name)`

test/cli_exec_test.go
```diff
@@ -143,3 +147,24 @@ func (suite *PouchExecSuite) TestExecExitCode(c *check.C) {
...
+	// give time for containerd restore, more than 3 second.
+	data, err := exec.Command("pidof", "containerd").CombinedOutput()
+	c.Assert(err, check.IsNil)
+	pid, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	c.Assert(err, check.IsNil)
+
+	syscall.Kill(pid, syscall.SIGTERM)
+	time.Sleep(5)
+
+	output := command.PouchRun("exec", name, "echo", "1").Stdout()
```
Letty5411: I would prefer to use:
```
res := command.PouchRun("exec", name, "echo", "1")
res.Assert(c, icmd.Success)
output := res.Stdout()
```

Ace-Tang:
@Letty5411 , from the test result, seems I kill containerd in test will influence cri test, maybe another containerd is need for the test.

ctrd/client.go
Ace-Tang:
```
 gopts := []grpc.DialOption{
        grpc.WithBlock(),
        grpc.WithInsecure(),
        grpc.WithTimeout(60 * time.Second),
        grpc.FailOnNonTempDialError(true),
        grpc.WithBackoffMaxDelay(3 * time.Second),
        grpc.WithDialer(dialer.Dialer),
    }
```
from `containerd.New` grpc connect get a 60s timeout, assumes that we do not consider about containerd killed unexpectly, valid of these connect should be consider.

PR #1483, opened 2018-06-07
bugfix: add attach volume when container start

rudyfly:
Add attach volume  when container start. Now volume attaching is
at container create stage, but if host have been restarted, container won't be
created, so volume can't be attached.

Add Config() for volume driver, it makes volume driver can use the
configure of volume in daemon. Now set `volume-meta-dir` and
`volume-timeout` into driver config.

NONE

A running container can start again after host have been restarted.

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -450,6 +450,8 @@ func (mgr *ContainerManager) Start(ctx context.Context, id, detachKeys string) (
 }
 
 func (mgr *ContainerManager) start(ctx context.Context, c *Container, detachKeys string) error {
+	var err error
+
 	c.Lock()
 	if c.Config == nil || c.State == nil {
 		c.Unlock()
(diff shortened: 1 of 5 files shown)


storage/volume/driver/driver_interface.go
```diff
@@ -26,6 +26,9 @@ type Driver interface {
 type Opt interface {
 	// Options return module customize volume options.
 	Options() map[string]types.Option
+
+	// Config is used to pass the daemon volume configure into driver.
+	Config(Context, map[string]interface{}) error
```
shaloulcy: I think the Driver interface should implement Config method, not the Opt interface. WDYT? @rudyfly

rudyfly: I have refacted it, PTAL 😄

daemon/mgr/container.go
```diff
@@ -450,18 +450,48 @@ func (mgr *ContainerManager) Start(ctx context.Context, id, detachKeys string) (
...
 	}
 	c.DetachKeys = detachKeys
 
+	// attach volume
+	for _, mp := range c.Mounts {
+		if mp.Name == "" {
+			continue
+		}
+
+		attachedVolumes := map[string]struct{}{}
```
yyb196: is it necessary to add a defer in a loop

rudyfly: oh, it should defer outside the loop

PR #1492, opened 2018-06-08
fix: set memory swap double of non-zero memory value

Ace-Tang:
set value of memory swap double of non-zero memory value if
not specify --memory-sawp.

2x memory swap value is advised but not must.

You should typically size your swap space to approximately 2x main
memory for systems with less than 2GB of RAM, or approximately 1x main
memory if you have more. If you do not have a lot of RAM, though, you
will generally want a lot more swap. It is not recommended that you
configure any less than 256M of swap on a system and you should keep in
mind future memory expansion when sizing the swap partition. The
kernel's VM paging algorithms are tuned to perform best when there is at
least 2x swap versus main memory. Configuring too little swap can lead
to inefficiencies in the VM page scanning code as well as create issues
later on if you add more memory to your machine. Finally, on larger
systems with multiple SCSI disks (or multiple IDE disks operating on
different controllers), we strongly recommend that you configure
swap on each drive. The swap partitions on the drives should be
approximately the same size. The kernel can handle arbitrary sizes but
internal data structures scale to 4 times the largest swap partition.
Keeping the swap partitions near the same size will allow the kernel to
optimally stripe swap space across the N disks. Do not worry about
overdoing it a little, swap space is the saving grace of UNIX and even
if you do not normally use much swap, it can give you more time to
recover from a runaway program before being forced to reboot.'

### DIFF
--- a/daemon/mgr/container_utils.go
+++ b/daemon/mgr/container_utils.go
@@ -228,6 +228,8 @@ func parsePSOutput(output []byte, pids []int) (*types.ContainerProcessList, erro
 
 // validateConfig validates container config
 func validateConfig(config *types.ContainerCreateConfig) ([]string, error) {
+	amendResource(&config.HostConfig.Resources)
+
 	// validates container hostconfig
 	warnings := make([]string, 0)
 	warns, err := validateResource(&config.HostConfig.Resources)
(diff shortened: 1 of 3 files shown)


Ace-Tang:
@sunyuan3 , this fix the aone bug.

sunyuan3:
LGTM

PR #1502, opened 2018-06-11
fix: support nullable bool value set in config

Ace-Tang:
make nullable value like `false` and 0 not be recognized as
empry value, in config set, they should be valid value.


rudyfly:
LGTM

PR #1521, opened 2018-06-13
refactor: return CRI services error

oiooj:
Refactor: We should not lose the error in CRI service.


YaoZengzeng:
LGTM

Thanks for your work @oiooj :)

PR #1523, opened 2018-06-14
bugfix: close error channel

ianwoolf:

function should close inside channle when return.

defer close(errChan)

### DIFF
--- a/cri/criservice.go
+++ b/cri/criservice.go
@@ -49,6 +49,7 @@ func runv1alpha1(daemonconfig *config.Config, containerMgr mgr.ContainerMgr, ima
 	}
 
 	errChan := make(chan error, 2)
+	defer close(errChan)
 	go func() {
 		errChan <- service.Serve()
 		logrus.Infof("CRI GRPC server stopped")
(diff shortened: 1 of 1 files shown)


cri/criservice.go
```diff
@@ -84,6 +85,7 @@ func runv1alpha2(daemonconfig *config.Config, containerMgr mgr.ContainerMgr, ima
 	}
 
 	errChan := make(chan error, 2)
+	defer close(errChan)
```
YaoZengzeng: The following situation will panic:

1. One goroutine send an error to errChan

2. The function return and the errChan will be closed

3. Another goroutine send an error to the closed errChan

Actually we don't need to close channel explicitly, unless we want to use it to notify receiver.

It will be GCed any way :)

ianwoolf: 1. the for loop will close only when is receive to error
3. the channel is the function inside channel, so no other goroutine will send to this channel

and if one channle is not close(), it would not be collected by GC() as long as the process in runninig.

YaoZengzeng: Please refer: https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open

ianwoolf: You are right. Unlike OS resource, the unclosed channel will be collected by GC.

oiooj: This is not related to GC actually, the Pouch should be shutdown when any error occured here, but there is a panic risk if we close the `errChan`.

YaoZengzeng: @oiooj can't agree more !

PR #1531, opened 2018-06-14
enhance: add volume lock

shaloulcy:

Right now,  the volume core has no lock. It may have problems in some race conditions. In this pull request, we add volume lock.

NONE


fuweid:
@shaloulcy Could you mind to provide more information about impact if we have the race condition?

shaloulcy:
@fuweid Fine. Now We create the volume with the same volume name. If the pouchd has no volume lock, the pouchd will create the volume in the same time, the result is unpredictable.

fuweid:
LGTM

fuweid:
LGTM

PR #1537, opened 2018-06-15
feature: add container setting check

Ace-Tang:
add container config settings check, include container config and
host config in create container, update container.
fix memory_swappiness initial value to zero, since -1 is invalid.

```
#echo -1 > /sys/fs/cgroup/memory/pouch/4d770c0e1a84320c513d9569511c0d65046eda456723340851356b9eb86c0e01/memory.swappiness 
bash: echo: write error: Invalid argument
```

### DIFF
--- a/daemon/mgr/container_utils.go
+++ b/daemon/mgr/container_utils.go
@@ -227,22 +227,23 @@ func parsePSOutput(output []byte, pids []int) (*types.ContainerProcessList, erro
 }
 
 // validateConfig validates container config
-func validateConfig(config *types.ContainerCreateConfig) ([]string, error) {
-	amendResource(&config.HostConfig.Resources)
-
+func validateConfig(config *types.ContainerConfig, hostConfig *types.HostConfig, update bool) ([]string, error) {
 	// validates container hostconfig
 	warnings := make([]string, 0)
-	warns, err := validateResource(&config.HostConfig.Resources)
+	warns, err := validateResource(&hostConfig.Resources, update)
 	if err != nil {
 		return nil, err
 	}
 	warnings = append(warnings, warns...)
 
+	if hostConfig.OomScoreAdj < -1000 || hostConfig.OomScoreAdj > 1000 {
+		return warnings, fmt.Errorf("oom score should be in range [-1000, 1000]")
+	}
 	// TODO: add more validate here
 	return warnings, nil
 }
 
-func validateResource(r *types.Resources) ([]string, error) {
+func validateResource(r *types.Resources, update bool) ([]string, error) {
 	cgroupInfo := system.NewCgroupInfo()
 	if cgroupInfo == nil {
 		return nil, nil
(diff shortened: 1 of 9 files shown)


daemon/mgr/container_utils.go
```diff
@@ -251,19 +252,25 @@ func validateResource(r *types.Resources) ([]string, error) {
...
+		if r.MemorySwap > 0 && !cgroupInfo.Memory.MemorySwap {
 			warn := "Current Kernel does not support memory swap, discard --memory-swap"
 			logrus.Warn(warn)
 			warnings = append(warnings, warn)
 			r.MemorySwap = 0
 		}
+		if r.Memory != 0 && r.Memory < MinMemory {
+			return warnings, fmt.Errorf("Minimal memory should greater than 4M")
+		}
+		if r.Memory == 0 && r.MemorySwap > 0 && !update {
```
fuweid: If it is the `create` action, the `--memory-swap` should also work with `memory`, right?

daemon/mgr/container_utils.go
```diff
@@ -251,19 +252,25 @@ func validateResource(r *types.Resources) ([]string, error) {
...
+		if r.Memory > 0 && !cgroupInfo.Memory.MemoryLimit {
 			warn := "Current Kernel does not support memory limit, discard --memory"
 			logrus.Warn(warn)
 			warnings = append(warnings, warn)
 			r.Memory = 0
 			r.MemorySwap = 0
 		}
-		if r.MemorySwap != 0 && !cgroupInfo.Memory.MemorySwap {
+		if r.MemorySwap > 0 && !cgroupInfo.Memory.MemorySwap {
 			warn := "Current Kernel does not support memory swap, discard --memory-swap"
```
fuweid: Could we put the `warn` as the const or package var? WDYT?

daemon/mgr/container_utils.go
```diff
@@ -307,12 +317,19 @@ func validateResource(r *types.Resources) ([]string, error) {
 			warnings = append(warnings, warn)
 			r.CPUQuota = 0
 		}
+		// cpu.cfs_quota_us can accept value less than 0, we allow -1 and > 1000
+		if r.CPUQuota > 0 && r.CPUQuota < 1000 {
+			return warnings, fmt.Errorf("CPU cfs quota should be greater than 1ms(1000)")
+		}
 		if r.CPUPeriod > 0 && !cgroupInfo.CPU.CPUPeriod {
 			warn := "Current Kernel does not support cpu period, discard --cpu-period"
```
fuweid: Could we put the warn as the const or package var? WDYT?

daemon/mgr/container_utils.go
```diff
@@ -307,12 +317,19 @@ func validateResource(r *types.Resources) ([]string, error) {
...
 			logrus.Warn(warn)
 			warnings = append(warnings, warn)
 			r.CPUPeriod = 0
 		}
+		if r.CPUPeriod != 0 && (r.CPUPeriod < 1000 || r.CPUPeriod > 1000000) {
```
fuweid: Since the number of `if` branch are too many, could we put the CPU related into one function? It can make the function readable....

test/cli_update_test.go
```diff
@@ -115,6 +115,32 @@ func (suite *PouchUpdateSuite) TestUpdateCpuPeriod(c *check.C) {
...
+// TestUpdateCpuMemoryFail is to verify the invalid value of updating container cpu and memory related flags will fail.
+func (suite *PouchUpdateSuite) TestUpdateCpuMemoryFail(c *check.C) {
+	name := "update-container-cpu-memory-period-fail"
+
+	res := command.PouchRun("run", "-d", "--name", name, busyboxImage, "top")
+	defer DelContainerForceMultyTime(c, name)
+	res.Assert(c, icmd.Success)
+
+	output := command.PouchRun("inspect", name).Stdout()
+	result := []types.ContainerJSON{}
```
fuweid: I don't think we need `result` here.  The `inspect` action should be tested by other suite.

daemon/mgr/container_utils.go
Ace-Tang: Yes, we check only in create, and memory-swap should work with memory

daemon/mgr/container_utils.go
Ace-Tang: Do you mean every warn make a const or warn ? will this too many?

test/cli_update_test.go
Ace-Tang: Yes, it can be removed.

fuweid:
LGTM. I think next step we should separate `validateResource` into several pieces to make it testable.

fuweid:
LGTM

PR #1542, opened 2018-06-15
feature: support shm size

Ace-Tang:

support run commond `pouch run -d --shm-size 1g  registry.hub.docker.com/library/busybox:latest top`

### DIFF
--- a/cli/container.go
+++ b/cli/container.go
@@ -71,6 +71,7 @@ type container struct {
 	cgroupParent   string
 	ulimit         Ulimit
 	pidsLimit      int64
+	shmSize        string
 
 	// log driver and log option
 	logDriver string
@@ -180,6 +181,10 @@ func (c *container) config() (*types.ContainerCreateConfig, error) {
 	if err != nil {
 		return nil, err
 	}
+	shmSize, err := opts.ParseShmSize(c.shmSize)
+	if err != nil {
+		return nil, err
+	}
 
 	config := &types.ContainerCreateConfig{
 		ContainerConfig: types.ContainerConfig{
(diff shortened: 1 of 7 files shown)


daemon/mgr/spec_mount.go
```diff
@@ -86,6 +91,10 @@ func setupMounts(ctx context.Context, c *Container, s *specs.Spec) error {
 
 		// TODO: support copy data.
 
+		if mp.Destination == "/dev/shm" && c.HostConfig.ShmSize != nil {
```
fuweid: Hi @Ace-Tang , I checked the swagger and found `If omitted, the system uses 64MB`. Is it correct for the API side? or it's just for the CLI?

cli/container.go
```diff
@@ -172,6 +173,11 @@ func (c *container) config() (*types.ContainerCreateConfig, error) {
 		return nil, err
 	}
 
+	shmSize, err := opts.ParseShmSize(c.shmSize)
```
fuweid: we are missing the `opts.ParseShmSize` function 😄

Ace-Tang: Oh, yes

daemon/mgr/spec_mount.go
Ace-Tang: I think the swagger readme is copy from something where, and 64m is not implement in code

Ace-Tang:
Oh, CI fails because of this case for many times.
```
```

daemon/mgr/spec_mount.go
fuweid: OK. So could you mind to help us to update the description? 😄

fuweid: yeah....containerd sets it to 64M. 

https://github.com/containerd/containerd/blob/master/oci/spec_unix.go#L129-L133

So, we keep it on the swagger.

fuweid:
LGTM

PR #1551, opened 2018-06-19
enhance: revise configure file to make travis happy

ZouRui89:

This pr changes ```echo "travis doesn't run for document-only changed"``` to ```echo travis doesn\'t run for document-only changed```, since ```"``` will cause the failure of json unmarsalling on the side of our friend pouchRobot.


Letty5411:
LGTM

PR #1576, opened 2018-06-22
feature: support net priority flag

Ace-Tang:
only implment net-priority in client end.

Only implement net-priority in client end, consider net resource control is not good enough, we not implement it in server client.

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -1964,6 +1964,10 @@ definitions:
       QuotaID:
         type: "string"
         description: "set disk quota by specified quota id, if id < 0, it means pouchd alloc a unique quota id"
+      NetPriority:
+        description: "net priority."
+        type: "integer"
+        default: 0
  
   ContainerCreateResp:
     description: "response returned by daemon when container create successfully"
--- a/cli/container.go
+++ b/cli/container.go
@@ -72,6 +72,7 @@ type container struct {
 	ulimit         Ulimit
 	pidsLimit      int64
 	shmSize        string
+	netPriority    int64
 
 	// log driver and log option
 	logDriver string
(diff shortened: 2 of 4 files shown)


apis/swagger.yml
```diff
@@ -1964,6 +1964,10 @@ definitions:
       QuotaID:
         type: "string"
         description: "set disk quota by specified quota id, if id < 0, it means pouchd alloc a unique quota id"
+      NetPriority:
+        description: "net priority."
```
allencloud: Could you add more description of this field? More detailed content makes more people know more about the great things you are doing.

apis/swagger.yml
```diff
@@ -1964,6 +1964,10 @@ definitions:
       QuotaID:
         type: "string"
         description: "set disk quota by specified quota id, if id < 0, it means pouchd alloc a unique quota id"
+      NetPriority:
```
shaloulcy: NetPriority seem a property of network endpoint. A container may have many endpoints, we may apply different Priorities to different endpoints. Can we consider the situation?

Ace-Tang: Since we not implement in sever, so we just add a flag called `NetPriority` to prepare

apis/swagger.yml
Ace-Tang: No, it will be `TODO` things.

cli/container.go
```diff
@@ -72,6 +72,7 @@ type container struct {
 	ulimit         Ulimit
 	pidsLimit      int64
 	shmSize        string
+	netPriority    int64
```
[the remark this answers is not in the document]
Ace-Tang: swagger generate with 
```
      type: "integer"
      default: 0
```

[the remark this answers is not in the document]
allencloud: If you wanna set a specific type, maybe you can consider the following way:
```
      BlkioWeight:
        description: "Block IO weight (relative weight), need CFQ IO Scheduler enable."
        type: "integer"
        format: "uint16"
        x-nullable: false
        minimum: 0
        maximum: 1000
```

rudyfly:
LGTM

PR #1431, opened 2018-05-29
feature: stats of cri manager

Starnop:
Signed-off-by: Starnop <starnop@163.com>

### Ⅰ. Describe what this PR did
This PR implement ContainerStats(), ListContainerStats() and ImageFsInfo() of CRI manager.

All of these interfaces are used to retrieve the metrics of containers or image file system info.

### Ⅱ. Does this pull request fix one issue?

None

### Ⅳ. Describe how to verify it
Use crictl tools:
```
# crictl stats
DEBU[0000] ListContainerStatsRequest: &ListContainerStatsRequest{Filter:&ContainerStatsFilter{Id:,PodSandboxId:,LabelSelector:map[string]string{},},} 

......

CONTAINER           CPU %               MEM                 DISK                INODES
0bbc13aa25d81       0.59                30.14MB             45.06kB             14
15eef79448135       0.00                286.7kB             24.58kB             9
2e257f09a2260       0.09                9.581MB             24.58kB             6
476ae4d251fe6       0.00                290.8kB             24.58kB             9
574b07cfd946c       0.03                10.21MB             8.192kB             3
587856b374711       0.44                11.49MB             24.58kB             10
6df66169f907a       1.18                37.26MB             61.44kB             19
725a1d78af2b0       0.00                290.8kB             24.58kB             9
8b44f5423bd64       0.02                6.955MB             12.29kB             4
a1f9ccae689ed       0.00                274.4kB             24.58kB             9
a42c9875b9e82       1.13                233.7MB             36.86kB             12
c4321c743307d       0.00                225.3kB             24.58kB             9
cf2dda929ea0b       0.00                286.7kB             24.58kB             9
eb400edd89bb5       0.00                5.947MB             28.67kB             9

```

or use curl tools:

```
# curl -X GET http://127.0.0.1:10255/stats/summary
```

PR #1577, opened 2018-06-22
feature: add systemd notify

### LINKED ISSUE #1568: [feature request] add systemd notify functionality
### Ⅰ. Issue Description

For the case that `systemctl start pouch`, we need to make sure that the pouchd has been started successfully. We cannot rely on the exit code of `systemctl` only,  we need to use `NOTIFY_SOCKET` to send the state to the systemd.

NOTE: https://github.com/coreos/go-systemd is third-party package which can help us to build this.

1.
2.
3.

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shaloulcy:
For the case that systemctl start pouch, we need to make sure that the
pouchd has been started successfully. We cannot rely on the exit code
of systemctl only, we need to use NOTIFY_SOCKET to send the state to
the systemd.

For the case that **systemctl start pouch**, we need to make sure that the pouchd has been started successfully. We cannot rely on the exit code of systemctl only, we need to use NOTIFY_SOCKET to send the state to the systemd.

So when the pouchd has set up , it will send the state to systemd

fixes #1568 

none

none

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -18,6 +18,8 @@ import (
 	"github.com/alibaba/pouch/pkg/meta"
 	"github.com/alibaba/pouch/pkg/system"
 
+	systemddaemon "github.com/coreos/go-systemd/daemon"
+	systemdutil "github.com/coreos/go-systemd/util"
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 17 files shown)


daemon/daemon.go
```diff
@@ -203,16 +205,30 @@ func (d *Daemon) Run() error {
...
 	}()
 
 	criStopCh := make(chan error)
-	go criservice.RunCriService(d.config, d.containerMgr, d.imageMgr, criStopCh)
+	go criservice.RunCriService(d.config, d.containerMgr, d.imageMgr, criStopCh, criReadyCh)
+
+	httpReady := <-httpReadyCh
+	criReady := <-criReadyCh
+
+	if httpReady && criReady {
```
yyb196: cri manager may don't start if `enable-cri=false` is specified, is it ok in this case?

shaloulcy: it's ok. If **enable-cri=false** is specified, we can read **true** from thehttpReadyCh

fuweid:
LGTM.

Since we don't have the adaptor to handle different servers, next step we can make it elegant.

fuweid:
But I need to confirm the version of dependency because it is different from `https://github.com/alibaba/pouch/pull/1431/files`

fuweid:
Confirm. We can merge it

Issue #1562, opened 2018-06-21
[feature request] pouch [ps|containers list API] with filter functionality


fuweid:
### Ⅰ. Issue Description

For now, the pouch list containers API doesn't support filter function. It's very useful for the monitor or schedule system. PouchContainer has the following requirements:

- [ ] id - Container’s ID
- [ ] name - Container’s 
- [ ] status - One of created, restarting, running, removing, paused, exited, or dead
- [ ] label - An arbitrary string representing either a key or a key-value pair. 

For example, we can use the following command to filter container by id:

```
pouch ps --filter id=fcd6279d2069
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yyb196:
filter according to label name and value is another requirment, moby only support filter by label euqals, we need filter by label doesn't equal. 
eg:
```
pouch ps --filter "label=color!=blue"
```

Ace-Tang:
Already support the follow:
- id - Container’s ID
- name - Container’s 
- status - One of created, restarting, running, removing, paused, exited, or dead
- label - An arbitrary string representing either a key or a key-value pair. 

also support unequal condition, like `pouch ps --filter "label=color!=blue"`

PR #1595, opened 2018-06-25
feature: support pouch ps filter

Ace-Tang:
support pouch ps --filter, not only equal condition, but also
unequal condition, like pouch ps --filter label=a!=b

Fixes: #1562.

fixes: #1562


daemon/mgr/container_list.go
```diff
@@ -0,0 +1,197 @@
...
+		case statusFilter:
+			match = fc.matchFilter(statusFilter, string(c.State.Status))
+		}
+
+		if match {
+			if name != statusFilter {
+				// consider if need to add non-running container
+				match = isNonStop || fc.all
+			}
+			break
```
yyb196: one match condition will stop this filter routine. what if people need both `-f label=site=daily` and `-f label=network!=overlay`

test/cli_ps_test.go
```diff
@@ -71,6 +71,124 @@ func (suite *PouchPsSuite) TestPsWorks(c *check.C) {
 
 }
 
+// TestPsFilterEqual tests "pouch ps -f" filter equal condition work
+func (suite *PouchPsSuite) TestPsFilterEqual(c *check.C) {
+	labelA := "equal-label-a"
+	command.PouchRun("run", "-d", "--name", labelA, "-l", "a=b", busyboxImage, "top").Assert(c, icmd.Success)
+	defer DelContainerForceMultyTime(c, labelA)
+	output := command.PouchRun("inspect", labelA).Stdout()
```
Letty5411: +	output := command.PouchRun("inspect", labelA).Assert(c,icmd.Success).Stdout()

test/cli_ps_test.go
```diff
@@ -71,6 +71,124 @@ func (suite *PouchPsSuite) TestPsWorks(c *check.C) {
...
+	_, exist1 = kv[labelA]
+	_, exist2 = kv[labelB]
+	_, exist3 = kv[name]
+	_, exist4 = kv[labelC]
+	c.Assert(exist1, check.Equals, false)
+	c.Assert(exist2, check.Equals, false)
+	c.Assert(exist3, check.Equals, true)
+	c.Assert(exist4, check.Equals, true)
+}
+
```
Letty5411: could also add negative tests, like:
-f  null
-f wrong format

and use multiple -f

daemon/mgr/container_list.go
Ace-Tang: Yes, one match will break loop, the function used to filter the match thing, so one condition match means it is chosed, 
it is what we want, we just want output matched thing.

test/cli_ps_test.go
Ace-Tang: All case has add.

daemon/mgr/container_list.go
Ace-Tang: Modify the logic, jump until condition not equal, and add flag `jump` to determine if a condition is useful. Test case has add for this case.

daemon/mgr/container_list.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if len(fc.condition) == 0 {
+		// if no filter condition, skip filter process,
+		// add container according to flag `all` or running status
+		return isNonStop || fc.all
+	}
+
+	// filter must meets all condition
+	var match, jump, statusKey = true, true, false
+	for name := range fc.condition {
```
fuweid: I have idea about this part.

The `jump` boolean means we should start next round check, right?  the `matchKVFilter` will return one boolean value if we run the `filters, exist := fc.condition[field]` before check.  Therefore, we don't need to check the `jump` in the following condition. 

Does it make senses to you?

test/cli_ps_test.go
```diff
@@ -71,6 +72,139 @@ func (suite *PouchPsSuite) TestPsWorks(c *check.C) {
 
 }
 
+// TestPsFilterInvalid tests "pouch ps -f" invalid
+func (suite *PouchPsSuite) TestPsFilterInvalid(c *check.C) {
+	result := command.PouchRun("ps", "-f", "foo")
```
fuweid: Next time, we can use the table-driven test to repeat the same thing.

test/cli_ps_test.go
```diff
@@ -71,6 +72,139 @@ func (suite *PouchPsSuite) TestPsWorks(c *check.C) {
...
+func (suite *PouchPsSuite) TestPsFilterEqual(c *check.C) {
+	labelA := "equal-label-a"
+	command.PouchRun("run", "-d", "--name", labelA, "-l", "a=b", busyboxImage, "top").Assert(c, icmd.Success)
+	defer DelContainerForceMultyTime(c, labelA)
+	output := command.PouchRun("inspect", labelA).Assert(c, icmd.Success).Stdout()
+	result := []types.ContainerJSON{}
+	if err := json.Unmarshal([]byte(output), &result); err != nil {
+		c.Errorf("failed to decode inspect output: %v", err)
+	}
+	labelAName := result[0].ID
```
fuweid: s/labelAName/labelAID/ maybe better

apis/swagger.yml
```diff
@@ -3529,6 +3529,33 @@ definitions:
...
 
+  ContainerListOptions:
+    description: |
+      options of list container, filters (a `map[string][]string`) to process on the container list. Available filters:
+
+      - `id=container-id`
+      - `name=container-name`
+      - `status=running`
+      - `label=key` or `label=key=value`
+      - `network=container-network`
```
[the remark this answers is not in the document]
Ace-Tang: I do not think so, interface should think more, we just tell which not support before this, user always not care about defined field

cli/ps.go
```diff
@@ -50,15 +52,26 @@ func (p *PsCommand) addFlags() {
...
+	filter, err := opts.ParseFilter(p.flagFilter)
+	if err != nil {
+		return err
+	}
+
 	var containers containerList
-	containers, err := apiClient.ContainerList(ctx, p.flagAll)
+
+	option := types.ContainerListOptions{All: p.flagAll}
+	if len(filter) > 0 {
```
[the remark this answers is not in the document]
Ace-Tang: No, less function call is better, `api.Client.ContainerList` has `len(filter) > 0` for robust.

client/container_list_test.go
```diff
@@ -57,7 +59,9 @@ func TestContainerList(t *testing.T) {
 	client := &APIClient{
 		HTTPCli: httpClient,
 	}
-	containers, err := client.ContainerList(context.Background(), true)
+	containers, err := client.ContainerList(context.Background(), types.ContainerListOptions{
+		All: true,
```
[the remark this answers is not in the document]
Ace-Tang: Construct `Container` struct is complete, let ci do this filter check.

daemon/mgr/container_list.go
```diff
@@ -0,0 +1,211 @@
...
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/alibaba/pouch/apis/opts"
+)
+
+var (
```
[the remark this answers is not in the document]
Ace-Tang: Not following you, could you explain more detail? @fuweid

daemon/mgr/container_list.go
```diff
@@ -0,0 +1,211 @@
...
+	return false, match
+}
+
+// filter does all select container work.
+func (fc *filterContext) filter(c *Container) bool {
+	isNonStop := c.IsRunningOrPaused()
+
+	// if filterFunc is defined, we not try other filter condition
+	if fc.filterFunc != nil {
+		if match := fc.filterFunc(c); match {
```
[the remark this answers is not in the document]
Ace-Tang: good

daemon/mgr/container_list.go
Ace-Tang: I also think about this, `filters, exist := fc.condition[field]` if field not exist, whether filter function return true or false is not make sense. Maybe need more think.

test/cli_ps_test.go
Ace-Tang: ^-^

test/cli_ps_test.go
Ace-Tang: This rule is for variable, WDYT?

apis/opts/filter.go
```diff
@@ -0,0 +1,105 @@
...
+	rawdata, err := json.Marshal(filter)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal filter params %v", filter)
+	}
+
+	return string(rawdata), nil
+}
+
+// FromURLParam gets filter from url query
+func FromURLParam(param string) (map[string][]string, error) {
```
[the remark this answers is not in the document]
Ace-Tang: Oh, good, I think about move the file `api/opts/filter.go` to `pkg` package and make it a package named filter, and I think the name `FromURLParam` and `ToURLParam` will make sense.

daemon/mgr/container_list.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if len(equalKV) == 0 && len(unequalKV) == 0 {
+		return false, false
+	}
+
+	match := false
+	for k, v := range value {
+		// filter equal condition
+		if equalValue, exist := equalKV[k]; exist {
+			if equalValue == "" {
```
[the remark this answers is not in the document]
Ace-Tang: good

client/container_list_test.go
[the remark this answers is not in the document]
fuweid: 😯 you are right.

apis/swagger.yml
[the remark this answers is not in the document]
fuweid: Solved

YaoZengzeng:
LGTM, it's all right for CRI part :)

PR #1597, opened 2018-06-26
refactor: move config file from cli into one places

Ace-Tang:
move config file cli/ulimit.go, cli/blkio.go, config/opt/runtime.go
to apis/opts/config, put all user defined flag var type into one places.


HusterWan:
LGTM

Issue #1602, opened 2018-06-27
[bug] SetStatus should set Running to true when started container


HusterWan:
### Ⅰ. Issue Description
we define `ContainerState` in `apis/types/container_state`, the `ContainerState` has a flag `Runnging`:

```
	Running bool `json:"Running,omitempty"`
```

we should set the `Running` flag to `true` when started container

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1604, opened 2018-06-27
bugfix: we should set Running flag to true when started container

HusterWan:

`SetStatus` set Running flag to true when started a container

fixes https://github.com/alibaba/pouch/issues/1602

### DIFF
--- a/daemon/mgr/container_state.go
+++ b/daemon/mgr/container_state.go
@@ -90,6 +90,7 @@ func (c *Container) SetStatusRunning(pid int64) {
 	c.State.StartedAt = time.Now().UTC().Format(utils.TimeLayout)
 	c.State.Pid = pid
 	c.State.ExitCode = 0
+	c.setStatusFlags(types.StatusRunning)
 }
 
 // SetStatusStopped sets a container to be status stopped.
@@ -107,6 +108,7 @@ func (c *Container) SetStatusStopped(exitCode int64, errMsg string) {
 	c.State.Pid = -1
 	c.State.ExitCode = exitCode
 	c.State.Error = errMsg
+	c.setStatusFlags(types.StatusStopped)
 }
 
 // SetStatusExited sets a container to be status exited.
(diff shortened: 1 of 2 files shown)


daemon/mgr/container_state.go
```diff
@@ -90,6 +90,7 @@ func (c *Container) SetStatusRunning(pid int64) {
 	c.State.StartedAt = time.Now().UTC().Format(utils.TimeLayout)
 	c.State.Pid = pid
 	c.State.ExitCode = 0
+	c.State.Running = true
```
Ace-Tang: Do we need to set `c.State.Running` to flase when container stop?

HusterWan: Good point, i will optimize soon

daemon/mgr/container_state.go
```diff
@@ -133,4 +137,33 @@ func (c *Container) SetStatusUnpaused() {
 	c.Lock()
 	defer c.Unlock()
 	c.State.Status = types.StatusRunning
+	c.setStatusFlags(types.StatusRunning)
+}
+
+// Notes(ziren): i still feel uncomfortable for a function hasing no return
+// setStatusFlags set the specified status flag to true, and unset others
+func (c *Container) setStatusFlags(status types.Status) {
+	statusFlags := map[types.Status]bool{
```
Ace-Tang: Oh， how about
```
 statusFlags := map[types.Status]bool{
        types.StatusDead:       true,
        types.StatusRunning:    true,
        types.StatusPaused:     true,
        types.StatusRestarting: true,
    }   

    if _, exists := statusFlags[status]; !exists {
        return
    }   

    c.State.Dead = true
```
do not be too concerned about this.

daemon/mgr/container_state.go
```diff
@@ -133,4 +137,33 @@ func (c *Container) SetStatusUnpaused() {
 	c.Lock()
 	defer c.Unlock()
 	c.State.Status = types.StatusRunning
+	c.setStatusFlags(types.StatusRunning)
```
Ace-Tang: I just wonder why use a function instead of a single line like `c.State.Running=true` ?

HusterWan: we have 4 flags : Dead, Paused, Restarting and Running , the flags are exclusive，so when set one flag, we should unset others. WDYT? @Ace-Tang

Ace-Tang:
LGTM

Issue #1151, opened 2018-04-17
[bug] exec with `-u nouser` hang


Letty5411:
### Ⅰ. Issue Description
`pouch exec -u admin` hangs when there is no user `admin`.

### Ⅱ. Describe what happened
```
#pouch run -d busybox sleep 100000e17ebb8a2f9487aa30b62982739e337b594029953dfc3f5bc0700ef0bcc5fecf


#pouch inspect e17ebb8a2f9487aa30b62982739e337b594029953dfc3f5bc0700ef0bcc5fecf|grep -i user
#pouch exec -u admin e17ebb8a2f9487aa30b62982739e337b594029953dfc3f5bc0700ef0bcc5fecf  dd if=/dev/zero of=/tmp/test/test/test bs=1024k count=1 &
[2] 35410

#pouch inspect e17ebb8a2f9487aa30b62982739e337b594029953dfc3f5bc0700ef0bcc5fecf|grep -i user
            "User": "admin"

#cat /proc/35410/stack
[<ffffffff810e1f65>] futex_wait_queue_me+0xc5/0x130
[<ffffffff810e2ae9>] futex_wait+0x179/0x280
[<ffffffff810e4b7e>] do_futex+0xfe/0x5b0
[<ffffffff810e50b0>] SyS_futex+0x80/0x180
[<ffffffff81646489>] system_call_fastpath+0x16/0x1b
[<ffffffffffffffff>] 0xffffffffffffffff

#pouch run -d busybox sleep 100000c7ca0c9cefaeb31e6ab9a975dc043bf999b6abe66383633544823b13cb0aafa5

#pouch exec c7ca0c9cefaeb31e6ab9a975dc043bf999b6abe66383633544823b13cb0aafa5 cat /etc/passwd
root:x:0:0:root:/root:/bin/sh
daemon:x:1:1:daemon:/usr/sbin:/bin/false
bin:x:2:2:bin:/bin:/bin/false
sys:x:3:3:sys:/dev:/bin/false
sync:x:4:100:sync:/bin:/bin/sync
mail:x:8:8:mail:/var/spool/mail:/bin/false
www-data:x:33:33:www-data:/var/www:/bin/false
operator:x:37:37:Operator:/var:/bin/false
nobody:x:65534:65534:nobody:/home:/bin/false
```
### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
[...]

PR #1605, opened 2018-06-28
bugfix: fix exec stuck when exec get error

Ace-Tang:
when daemon deal with exec process get error, process will get stuck,
since http connection has been hijacked, it won't finish and return
with error message. Fix it by write error message into hijacked
connection, close io when process get error.

fixes: #1151

fixes: #1151

### DIFF
--- a/apis/server/exec_bridge.go
+++ b/apis/server/exec_bridge.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
 )
 
 func (s *Server) createContainerExec(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
@@ -55,8 +56,13 @@ func (s *Server) startContainerExec(ctx context.Context, rw http.ResponseWriter,
 	name := mux.Vars(req)["name"]
 	_, upgrade := req.Header["Upgrade"]
 
+	if err := s.ContainerMgr.CheckExecExist(ctx, name); err != nil {
+		return err
+	}
+
 	var attach *mgr.AttachConfig
 
+	// TODO(huamin.thm): support detach exec process through http post method
 	if !config.Detach {
 		hijacker, ok := rw.(http.Hijacker)
 		if !ok {
@@ -72,7 +78,11 @@ func (s *Server) startContainerExec(ctx context.Context, rw http.ResponseWriter,
 		}
 	}
 
-	return s.ContainerMgr.StartExec(ctx, name, config, attach)
+	if err := s.ContainerMgr.StartExec(ctx, name, attach); err != nil {
+		logrus.Errorf("failed to run exec process: %s", err)
+	}
+
+	return nil
 }
 
 func (s *Server) getExecInfo(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
--- a/apis/server/router.go
+++ b/apis/server/router.go
@@ -186,6 +186,7 @@ func filter(handler handler, s *Server) http.HandlerFunc {
 			return
 		}
 		// Handle error if request handling fails.
+		logrus.Errorf("Handler for %s %s, client %s returns error: %s", req.Method, req.URL.RequestURI(), clientInfo, err)
 		HandleErrorResponse(w, err)
 	}
 }
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -1,15 +1,17 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"os"
-	"sync"
 
 	"github.com/alibaba/pouch/apis/types"
-	"github.com/docker/docker/pkg/stdcopy"
 
+	"github.com/docker/docker/pkg/stdcopy"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -61,6 +63,7 @@ func (e *ExecCommand) runExec(args []string) error {
 	id := args[0]
 	command := args[1:]
 
+	// TODO(huamin.thm): exec detach not implement now, detach mode not hijack connect
 	createExecConfig := &types.ExecCreateConfig{
 		Cmd:          command,
 		Tty:          e.Terminal,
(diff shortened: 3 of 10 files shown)


Ace-Tang:
Hi, @Letty5411 , I want you to help look at a test
```
func (suite *APIContainerExecStartSuite) TestContainerExecStartNotFound(c *check.C) {
    obj := map[string]interface{}{
        "Detach": false,
        "Tty":    false,
    }   
    body := request.WithJSONBody(obj)
    resp, err := request.Post("/exec/TestContainerExecStartNotFound/start", body)
    c.Assert(err, check.IsNil)
    CheckRespStatus(c, resp, 404)
}
```
As for fix exec hang, I ignore the error `https://github.com/alibaba/pouch/pull/1605/files#diff-b7f6767e7c90c31b2aec9ef134e49e54L75`, since error message can not be return by a hijacked connect. Does the test fails because I ignore the error?

daemon/mgr/container_exec.go
```diff
@@ -3,11 +3,13 @@ package mgr
 import (
 	"context"
 	"fmt"
+	"io"
 
 	"github.com/alibaba/pouch/apis/types"
 	"github.com/alibaba/pouch/ctrd"
 	"github.com/alibaba/pouch/pkg/errtypes"
 	"github.com/alibaba/pouch/pkg/randomid"
+	"github.com/docker/docker/pkg/stdcopy"
```
fuweid: move the package into next group

cli/exec.go
```diff
@@ -85,53 +88,81 @@ func (e *ExecCommand) runExec(args []string) error {
...
 
-	execInfo, err := apiClient.ContainerExecInspect(ctx, createResp.ID)
-	if err != nil {
-		return err
-	}
+	select {
+	case err := <-stdoutDone:
+		if err != nil {
+			logrus.Debugf("receive stdout error: %s", err)
+			return err
```
fuweid: If the code goes into this part, the following code will go into the end?

```
go func() {
    if stdin {
       io.Copy(conn, os.Stdin)
    }

    close(stdinDone)
}()
```

fuweid: It is nit comment. If there exists goroutine leak, the cli will die to make the goroutine exit....

daemon/mgr/container_exec.go
Ace-Tang: Oh!

cli/exec.go
Ace-Tang: Maybe depend on when golang schedule the goroutine if stdin is `flase`, if stdin if true, it must became a leak.

Ace-Tang:
@YaoZengzeng , Could you help to see if the patch will infect CRI part.

cli/exec.go
fuweid: Yes. But the cli will return and exit so that the leak problem will be solved. LGTM right now.

daemon/mgr/container_exec.go
```diff
@@ -86,19 +102,11 @@ func (mgr *ContainerManager) StartExec(ctx context.Context, execid string, confi
 	c.Unlock()
 
 	execConfig.Running = true
-	defer func() {
-		if err != nil {
-			execConfig.Running = false
-			exitCode := 126
-			execConfig.ExitCode = int64(exitCode)
-		}
-		mgr.ExecProcesses.Put(execid, execConfig)
```
Ace-Tang: I think `mgr.ExecProcesses.Put(execid, execConfig)` is unnecessary， since it has added in `CreateExec`, WDYT, @fuweid

daemon/mgr/container_exec.go
```diff
@@ -37,25 +39,39 @@ func (mgr *ContainerManager) CreateExec(ctx context.Context, name string, config
...
+
+	eio, err := mgr.openExecIO(execid, attach)
+	if err != nil {
+		return err
 	}
 
+	defer func() {
+		if err != nil {
+			var stdout io.Writer = eio.Stdout
+			if !execConfig.Tty && !eio.MuxDisabled {
```
Ace-Tang: update the judgement, @fuweid

daemon/mgr/container_exec.go
Ace-Tang: @allencloud , @HusterWan , does `mgr.ExecProcesses.Put(execid, execConfig)` can be removed here?

HusterWan: @Ace-Tang I do not think so, we will use `exec/EXEC_ID/json` to judge the exec result, so we should update the `execConfig` when exec done

cli/exec.go
```diff
@@ -1,14 +1,16 @@
...
+	"bufio"
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"os"
-	"sync"
 
 	"github.com/alibaba/pouch/apis/types"
 	"github.com/docker/docker/pkg/stdcopy"
```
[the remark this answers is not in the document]
Ace-Tang: I will try...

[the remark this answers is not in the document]
Ace-Tang: But I think this part is not fixed time time.

Ace-Tang:
Updated as the comment, @HusterWan @fuweid

apis/server/exec_bridge.go
```diff
@@ -55,8 +56,13 @@ func (s *Server) startContainerExec(ctx context.Context, rw http.ResponseWriter,
 	name := mux.Vars(req)["name"]
 	_, upgrade := req.Header["Upgrade"]
 
+	if err := s.ContainerMgr.CheckExecExist(ctx, name); err != nil {
```
allencloud: I am wondering if we should keep consistent that object existence checking should be located in the manager layer rather than bridge layer?

So maybe we should encapsulate this checking into the function `StartExec`.

apis/server/router.go
```diff
@@ -186,6 +186,7 @@ func filter(handler handler, s *Server) http.HandlerFunc {
 			return
 		}
 		// Handle error if request handling fails.
+		logrus.Errorf("Handler for %s %s, client %s returen error: %s", req.Method, req.URL.RequestURI(), clientInfo, err)
```
allencloud: s/returen/returned

apis/server/exec_bridge.go
Ace-Tang: if check in `StartExec`, it too late. `StartExec` will not return error, since hikacked connection not take error back, we only write error message into connect, then api will get the error message, if check is in StartExec, we will never get error.

fuweid:
LGTM

PR #1614, opened 2018-07-02
bugfix: fix memory-swap flag not validate correct

Ace-Tang:
amendContainerSettings should before validateConfig


Letty5411:
LGTM

PR #1621, opened 2018-07-04
bugfix: release the container resources if contaienr failed to start

shaloulcy:

When the container failed to start, we should release the container resources, including network and containerio.

NONE

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -444,6 +444,14 @@ func (mgr *ContainerManager) start(ctx context.Context, c *Container, detachKeys
 		if err == nil {
 			return
 		}
+
+		// release the container resources(network and containerio)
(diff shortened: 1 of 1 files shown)


daemon/mgr/container.go
```diff
@@ -444,6 +444,14 @@ func (mgr *ContainerManager) start(ctx context.Context, c *Container, detachKeys
 		if err == nil {
 			return
 		}
+
+		// release the container resources(network and containerio)
+		err = mgr.releaseContainerResources(c)
```
HusterWan: if failed before prepareContainerNetwork, will already occur error here. WDTY??

shaloulcy: Yes, releaseContainerResources will still work. It will check the existence of network.

HusterWan:
lgtm

PR #1622, opened 2018-07-04
bugfix: execute remount rootfs before prestart hook

HusterWan:

when execute prestart hook, if container's rootfs not mounted will cause start failed. so, we should mount the rootfs before excute prestart hook.


rudyfly:
LGTM

PR #1647, opened 2018-07-06
feature: deamon support --log-driver and --log-opt options

### LINKED ISSUE #1646: [feature request] pouch daemon starts with default --log-driver and --log-opt options
K8S scheduler can use CRI to connect with PouchContainer. However, the CRI doesn't define the log configuration in the request for now. 

For the swarm user, they maybe need to configure the log-driver and log-opt for the container. For this case, we should allow the user to setup up the default configuration in daemon side so that we don't need to do the configuration in CRI.

Therefore, pouchd should can start with --log-driver and --log-opt, like

```
pouchd --log-driver syslog ...
```

zhuangqh:

pouchd could start with --log-driver and --log-opt as the default configuration of log driver.

fixes #1646 

Due to the defect of go-swagger v0.12, the validation of these options are temporarily noneffective.

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -309,11 +309,7 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 	}
 
 	// set default log driver and validate for logger driver
-	if config.HostConfig.LogConfig == nil {
-		config.HostConfig.LogConfig = &types.LogConfig{
-			LogDriver: types.LogConfigLogDriverJSONFile,
-		}
-	}
+	config.HostConfig.LogConfig = mgr.getDefaultLogConfigIfMissing(config.HostConfig.LogConfig)
 
 	container := &Container{
 		State: &types.ContainerState{
@@ -412,6 +408,29 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 		Name:     name,
 		Warnings: warnings,
 	}, nil
+}
+
+func (mgr *ContainerManager) getDefaultLogConfigIfMissing(logConfig *types.LogConfig) *types.LogConfig {
+	defaultLogOpts := make(map[string]string)
+	for k, v := range mgr.Config.DefaultLogConfig.LogOpts {
+		defaultLogOpts[k] = v
+	}
+
+	if logConfig == nil {
+		defaultConfig := mgr.Config.DefaultLogConfig
+		defaultConfig.LogOpts = defaultLogOpts
+		return &defaultConfig
+	}
+
+	if logConfig.LogDriver == "" {
+		logConfig.LogDriver = mgr.Config.DefaultLogConfig.LogDriver
+	}
+
+	if len(logConfig.LogOpts) == 0 {
+		logConfig.LogOpts = defaultLogOpts
+	}
+
+	return logConfig
 }
 
 // Get the detailed information of container.
--- a/main.go
+++ b/main.go
@@ -9,7 +9,9 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/alibaba/pouch/apis/opts"
 	optscfg "github.com/alibaba/pouch/apis/opts/config"
+	"github.com/alibaba/pouch/apis/types"
 	"github.com/alibaba/pouch/daemon"
 	"github.com/alibaba/pouch/daemon/config"
 	"github.com/alibaba/pouch/lxcfs"
@@ -29,6 +31,7 @@ import (
 var (
 	sigHandles   []func() error
 	printVersion bool
+	logOpts      []string
 )
 
 var cfg = &config.Config{}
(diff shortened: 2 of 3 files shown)


zhuangqh:
PTAL @fuweid

main.go
```diff
@@ -105,6 +107,10 @@ func setupFlags(cmd *cobra.Command) {
 	flagSet.BoolVar(&cfg.NetworkConfig.BridgeConfig.IPForward, "ipforward", true, "Enable ipforward")
 	flagSet.BoolVar(&cfg.NetworkConfig.BridgeConfig.UserlandProxy, "userland-proxy", false, "Enable userland proxy")
 
+	// log config
+	flagSet.StringVar(&cfg.DefaultLogConfig.LogDriver, "log-driver", "", "Set default log driver")
```
fuweid: I think we can set the `log-driver` to `jsonfile`, like mody does. WDTY?

main.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/alibaba/pouch/storage/quota"
 	"github.com/alibaba/pouch/version"
 
+	"github.com/alibaba/pouch/apis/opts"
```
fuweid: please put this line to previous group.

main.go
zhuangqh: yep. If `log-driver` is empty, CI will fail.

daemon/mgr/container.go
```diff
@@ -310,9 +310,8 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 
 	// set default log driver and validate for logger driver
 	if config.HostConfig.LogConfig == nil {
-		config.HostConfig.LogConfig = &types.LogConfig{
-			LogDriver: types.LogConfigLogDriverJSONFile,
-		}
+		defaultConfig := mgr.Config.DefaultLogConfig
+		config.HostConfig.LogConfig = &defaultConfig
```
fuweid: The request maybe come from swarm. If it does, the `config.HostConfig.LogConfig` is not nil but the value is empty. For this case, the `config.HostConfig.LogConfig.LogDriver` is empty and `config.HostConfig.LogConfig. LogOpts` is nil. 

It isn't good practice in API design. However, we need to compatible with legacy system. We need to add other check that  `config.HostConfig.LogConfig` is empty or not. WDTY?

daemon/mgr/container.go
```diff
@@ -309,10 +309,11 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 	}
 
 	// set default log driver and validate for logger driver
-	if config.HostConfig.LogConfig == nil {
-		config.HostConfig.LogConfig = &types.LogConfig{
-			LogDriver: types.LogConfigLogDriverJSONFile,
-		}
+	// note: request from swarm is not standard. config.HostConfig.LogConfig is not nil, but its value is empty.
+	//       We need to be compatible with this legacy system.
+	if config.HostConfig.LogConfig == nil || config.HostConfig.LogConfig.LogDriver == "" {
```
fuweid: Could we use helper function to check the `config.HostConfig.LogConfig` is empty or not? 

Like, 

```
func isEmptyLogConfig(xx *types.LogConfig) bool {
      if config.HostConfig.LogConfig == nil  {
           return true
      }
      return config.HostConfig.LogConfig.LogDriver == "" && len(config.HostConfig.LogConfig.LogOpts) == 0
}
```

WDTY?

main.go
```diff
@@ -105,6 +107,10 @@ func setupFlags(cmd *cobra.Command) {
 	flagSet.BoolVar(&cfg.NetworkConfig.BridgeConfig.IPForward, "ipforward", true, "Enable ipforward")
 	flagSet.BoolVar(&cfg.NetworkConfig.BridgeConfig.UserlandProxy, "userland-proxy", false, "Enable userland proxy")
 
+	// log config
+	flagSet.StringVar(&cfg.DefaultLogConfig.LogDriver, "log-driver", "jsonfile", "Set default log driver")
```
fuweid: s/"jsonfile"/types.LogConfigLogDriverJSONFile 😄

daemon/mgr/container.go
zhuangqh: Yep. Thanks

daemon/mgr/container.go
```diff
@@ -414,6 +410,23 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 	}, nil
 }
 
+// set manager log config to the container's log config if the container's log config is not specified.
+func (mgr *ContainerManager) setDefaultLogConfig(logConfig *types.LogConfig) {
```
fuweid: s/setDefaultLogConfig/setDefaultLogConfigIfMissing/g

daemon/mgr/container.go
```diff
@@ -414,6 +410,23 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
...
+	if logConfig == nil {
```
fuweid: I don't think we can set the value to the config, since it is not pointer to pointer

fuweid:
LGTM

Issue #1640, opened 2018-07-05
[help wanted] add unit-test for proxy/canonicalAddr


zhuangqh:
### Ⅰ. Issue Description
Add unit-test for canonicalAddr method which locate on `ctrd/image_proxy_util.go`.

You can take [env_test.go](https://github.com/alibaba/pouch/blob/master/apis/opts/env_test.go) for reference.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅴ. Anything else we need to know?
please "Baiji" developer refer to this issue in your PR for the convenience for us to count your score.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Issue #1638, opened 2018-07-05
[help wanted] add unit-test for ValidateCPUQuota


zhuangqh:
### Ⅰ. Issue Description
Add unit-test for ValidateCPUQuota method which locate on `apis/opts/cpu.go`.

You can take [env_test.go](https://github.com/alibaba/pouch/blob/master/apis/opts/env_test.go) for reference.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅴ. Anything else we need to know?
please "Baiji" developer refer to this issue in your PR for the convenience for us to count your score.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Issue #1637, opened 2018-07-05
[help wanted] add unit-test for ValidateCPUPeriod


zhuangqh:
### Ⅰ. Issue Description
Add unit-test for ValidateCPUPeriod method which locate on `apis/opts/cpu.go`.

You can take [env_test.go](https://github.com/alibaba/pouch/blob/master/apis/opts/env_test.go) for reference.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅴ. Anything else we need to know?
please "Baiji" developer refer to this issue in your PR for the convenience for us to count your score.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zhuangqh:
Why is there no `areas/test` label? I think the description of this issue is similar to #1638 
please cc @ZouRui89

ZouRui89:
Checked the logics of add-labels-to-issue part in cute robot...
This situation should not have happened. I think something might have stuck the label event. 😢

PR #1702, opened 2018-07-10
wangyifei24  5组  issue:#1637 #1638 #1639 #1640

wangyifei24:


wangyifei24:
wangyifei24 5组 issue:#1637 #1638 #1639 #1640

PR #1724, opened 2018-07-12
enhance: add .DS_Store ignore

Frankzhaopku:
Add .DS_Store file ignore pattern in .gitignore to prevent contributor commit these files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage.txt
 .editorconfig
 .bashrc
 .idea
+.DS_Store


.gitignore
```diff
@@ -8,3 +8,4 @@ coverage.txt
 .editorconfig
 .bashrc
 .idea
+.DS_Store
```
fuweid: `*/.DS_Store` is better for any dir in the root. WDTY?

Frankzhaopku: @fuweid Actually .DS_Store already ignored all the files in sub directories, if this is what you mean.

fuweid: 👌 LGTM

PR #1729, opened 2018-07-12
bugfix: Fix log-opt option parse fails if value contains comma

Frankzhaopku:
pflag.StringSliceVar will split the value into string array with
comma, instead pflag.StringArrayVar will pass the value as a whole
string, so we should use StringArrayVar if we intend to pass the
value into other component.

#1715


fuweid:
LGTM

fuweid:
LGTM

fuweid:
fix #1715

fuweid:
close #1715

PR #1731, opened 2018-07-13
build: change version from 0.5.0 to 1.0.0-rc1

fuweid:

change version from 0.5.0 to 1.0.0-rc1

NONE

NONE

NONE

### DIFF
--- a/hack/build
+++ b/hack/build
@@ -8,7 +8,7 @@ GOPATH=$BUILDPATH
 
 
 # Go parameters
-VERSION="0.5.0"
+VERSION="1.0.0-rc1"
 APIVERSION="1.24"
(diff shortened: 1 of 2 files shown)


hack/build
```diff
@@ -8,7 +8,7 @@ GOPATH=$BUILDPATH
 
 
 # Go parameters
-VERSION="0.5.0"
+VERSION="1.0.0-rc1"
```
HusterWan: how about 1.0.0-rc1-dev ?  after deadline to release, then we change the version to 1.0.0-rc1

HusterWan:
lgtm

hack/build
fuweid: since we will use rc2, we don't need dev in the version 😄

PR #1743, opened 2018-07-17
fix: ensure daemon start successfully after update damon offline

Ace-Tang:

1. make Namespace field not marshal to config file to ensure daemon
restart successful after update daemon offline.
2. add update daemon offline test.
3. fix update daemon error when config file is not exist.


rudyfly:
LGTM

PR #1750, opened 2018-07-18
refactor: reorder hack scripts

### LINKED ISSUE #1582: [help wanted] reorder hack folder to make script easy to maintain
### Ⅰ. Issue Description

The `hack` folder contains a lot of shell script. However, there is a little mess in the script. We should reorder or refactor the hack folder. 

For the PouchContainer, both CI and Makefile should share the same script from hack if they do the same things. Functionally, there are there parts:

* lint or verify: check code style and typo in the repo
* build: build image, go binary, rpm/deb packages...
* install dep: install anythings we need

> NOTE: We can check the DEP is existing or not before downloading. Save time

Those scripts should be shared by other module.

This change not just help to make code clear, but also help to improve our docs.

1.
2.
3.

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

fuweid:

Reorder hack scripts and make the code coverage correct.

fixes #1582 

* use `hack/install` dir to contains all about the dependencies install thing
* use `hack/testing` dir to contains all about the integration thing.
* use gometalinter to do all the linters

However, I still maintain the `hack/build` and `hack/make.sh` since both the `dep` and `rpm` packages need them. In the following the PR, we should remove the legacy code and refactor the package-scripts.

Besides this, I also refactor `travis.yml` to splits the testing into three jobs.

* Unit-Test
* Integration-Test
* CRI-Integration-Test

Based on this, we can use the tag for the job coverage so that `Codecov` can combine three kinds of coverages into one.  Please check the comment created by `Codecov`, like:

| Flag | Coverage Δ | |
|---|---|---|
| #critest | `33.52% <ø> (?)` | |
| #integrationtest | `37.67% <ø> (?)` | |
| #unittest | `20.33% <ø> (?)` | |

Eye ball?


sunyuan3:
LGTM

fuweid:
The change is very large. @YaoZengzeng , @Letty5411 and @chuanchang please take a look again. Thanks

hack/testing/run_daemon_integration.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+source utils.sh
+
```
Letty5411: I would prefer to use absolute dir instead of `cd`  to source the utils.sh

Letty5411:
Another question is do we need to modify related docs in pouch/doc?

PR #1853, opened 2018-07-24
add unit-test for DetachVolume team4

SeaRise:

fixes 1761

### DIFF
--- a/storage/volume/core_test.go
+++ b/storage/volume/core_test.go
@@ -229,5 +229,40 @@ func TestAttachVolume(t *testing.T) {
 }
 
 func TestDetachVolume(t *testing.T) {
-	// TODO
+	volName1 := "vol2"
+	driverName1 := "fake_driver12"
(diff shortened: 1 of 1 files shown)


storage/volume/core_test.go
```diff
@@ -229,5 +229,40 @@ func TestAttachVolume(t *testing.T) {
...
+		t.Fatalf("create volume error: %v", err1)
+	}
+	if v1.Name != volName1 {
+		t.Fatalf("expect volume name is %s, but got %s", volName1, v1.Name)
+	}
+	if v1.Driver() != driverName1 {
+		t.Fatalf("expect volume driver is %s, but got %s", driverName1, v1.Driver())
+	}
+
+	v2, err2 := core.DetachVolume(volid1, nil)
```
HusterWan: this test case is ok , but we can do better.

how about add more situations like attach the volume to a container, then do DetachVolume action.

PR #1824, opened 2018-07-24
fix issue #1761 baiji-group 264-2

panzaiyu:
### Ⅰ. Describe what this PR did

Add unit-test for Volume Core's DetachVolume method
### Ⅱ. Does this pull request fix one issue?

fixes #1761

PR #1877, opened 2018-07-24
Issue6 1

### LINKED ISSUE #1761: [help wanted] add unit-test for DetachVolume
### Ⅰ. Issue Description
Add unit-test for Volume Core's DetachVolume method which locate on storage/volume/core.go.

1.
2.
3.

please "Baiji" developer refer to this issue in your PR for the convenience for us to count your score.

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

YitianChen1996:
TestDetachVolume

fix #1761

I add some code for testing method Core.DetachVolume
I make two test cases in this issue
case 1 is for testing condition that detach volume not exists or not attached
case 2 is for testing condition that detach an attached volume 

the run Test cases an it can pass

In my work, I found that the parameter string[] labels is never used in the code, I suspect that is a bug

### DIFF
--- a/ctrd/image_proxy_util_test.go
+++ b/ctrd/image_proxy_util_test.go
@@ -40,7 +40,7 @@ func TestCanonicalAddr(t *testing.T) {
 
 func TestUseProxy(t *testing.T) {
 	type args struct {
-		str string
+		str     string
 		noProxy string
 	}
 	tests := []struct {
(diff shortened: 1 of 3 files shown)


YaoZengzeng:
#1824 which has been merged solves the same problem with this PR, so close it.

Thanks for your work !

PR #1880, opened 2018-07-24
Test1

zhengshiqiang47:

#1772 

baiji group five


JeromeTan1997:
The scenario that the value of environmental variable `no_proxy` is `*` is not tested. @Lykosteng

PR #1926, opened 2018-07-30
bugfix: don't set mount tab when rootfs diskquota is null

rudyfly:
Fixes: When rootfs diskquota is null, it will not set mount tab

NONE


CodeJuan:
LGTM

PR #2064, opened 2018-08-07
refactor: move functions from pre-start hook to pouchd

rudyfly:
move the functions: `set rootfs quota` and `set mtab` from pre-start
hook in runc to pouchd container's create. Pouchd mounts the rootfs and
do these functions, if it is successful, and then umount the rootfs.


daemon/mgr/container.go
```diff
@@ -2224,6 +2236,104 @@ func (mgr *ContainerManager) populateVolumes(ctx context.Context, c *Container)
 	return nil
 }
 
+func (mgr *ContainerManager) setMountTab(ctx context.Context, c *Container) error {
```
HusterWan: `mgr/container.go` are too big,  some functions are worked for container object, maybe we can put the code into `container_type.go`, WDYT???

rudyfly: I add a new file `container_storage.go` to manager the operation of container storage module.

Ace-Tang:
@rudyfly , get the ci passed, then LTGM, except a question: would mount in pouchd meet the `mount too many layers caused over page limit` problem.

PR #2073, opened 2018-08-10
bugfix: avoid the deadlock when failed to remove invalid sandbox

YaoZengzeng:

It has the possibility that we failed to run the sandbox and it is not being cleaned up.

Kubelet will use list to get the sandboxes, but will not get the status of the failed pod whose meta data has not been put into the Sandbox Store.

And Kubelet will keep trying to get the status of the failed pod and won't create a new one to replace it. 

It's a DEAD LOCK.

Actually Kubelet should not know the existent of invalid pod whose meta data won't be in the Sandbox Store.

So we could avoid the DEAD LOCK mentioned above.

### DIFF
--- a/cri/v1alpha1/cri.go
+++ b/cri/v1alpha1/cri.go
@@ -207,7 +207,10 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 	defer func() {
 		// If running sandbox failed, clean up the container.
 		if retErr != nil {
-			c.ContainerMgr.Remove(ctx, id, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true})
+			err := c.ContainerMgr.Remove(ctx, id, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true})
+			if err != nil {
+				logrus.Errorf("failed to remove the container when running sandbox failed: %v", err)
+			}
 		}
 	}()
 
@@ -460,6 +463,11 @@ func (c *CriManager) ListPodSandbox(ctx context.Context, r *runtime.ListPodSandb
 	sandboxList, err := c.ContainerMgr.List(ctx, opts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list sandbox: %v", err)
+	}
+
+	sandboxList, err = c.filterInvalidSandboxes(ctx, sandboxList)
+	if err != nil {
+		return nil, fmt.Errorf("failed to filter invalid sandboxes: %v", err)
 	}
 
 	sandboxes := make([]*runtime.PodSandbox, 0, len(sandboxList))
--- a/cri/v1alpha1/cri_utils.go
+++ b/cri/v1alpha1/cri_utils.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containerd/typeurl"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	"github.com/go-openapi/strfmt"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"golang.org/x/sys/unix"
 	"k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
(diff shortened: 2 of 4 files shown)


cri/v1alpha1/cri_utils.go
```diff
@@ -313,6 +313,43 @@ func toCriSandbox(c *mgr.Container) (*runtime.PodSandbox, error) {
 	}, nil
 }
 
+// It has the possibility that we failed to run the sandbox and it is not being cleaned up.
+// Kubelet will use list to get the sandboxes, but will not get the status of the failed pod
+// whose meta data has not been put into the Sandbox Store. And Kubelet will keep trying to
+// get the status of the failed pod and won't create a new one to replace it. It's a DEAD LOCK.
+// Actually Kubelet should not know the existent of invalid pod whose meta data won't be in the
```
HusterWan: %s/existent/existence ?

cri/v1alpha1/cri_utils.go
```diff
@@ -313,6 +313,43 @@ func toCriSandbox(c *mgr.Container) (*runtime.PodSandbox, error) {
...
+			result = append(result, sandbox)
+			continue
+		}
+
+		status := sandbox.State.Status
+		// NOTE: what if the worst case that we failed to remove the sandbox and
+		// it is still running?
+		if status != apitypes.StatusRunning && status != apitypes.StatusCreated {
+			// Remove invalid sandbox.
+			c.ContainerMgr.Remove(ctx, sandbox.ID, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true})
```
HusterWan: Remove is a dangerous action, should we just print a warning log and just keep the invalid containers.

Maybe some containers are invalid for cri, but valid for pouchd, WDYT?

YaoZengzeng: The containers belong to pouchd have been filtered already.

The containers here are CRI sandbox.

If we don't remove the invalid sandbox container here, they will never be removed :)

rudyfly:
I think add checking invalid sandboxes is necessary, but we should try to know why remove container is failed？

YaoZengzeng:
@rudyfly Yes, maybe we should print the error message when failed to remove the invalid sandbox.

cri/v1alpha1/cri_utils.go
```diff
@@ -313,6 +313,43 @@ func toCriSandbox(c *mgr.Container) (*runtime.PodSandbox, error) {
...
+// Actually Kubelet should not know the existence of invalid pod whose meta data won't be in the
+// Sandbox Store. So we could avoid the DEAD LOCK mentioned above.
+func (c *CriManager) filterInvalidSandboxes(ctx context.Context, sandboxes []*mgr.Container) ([]*mgr.Container, error) {
+	validSandboxes, err := c.SandboxStore.Keys()
+	if err != nil {
+		return nil, err
+	}
+
+	var result []*mgr.Container
+	for _, sandbox := range sandboxes {
```
Starnop: How about put this layer loop outside ? :) 
Expect that, LGTM.

cri/v1alpha1/cri.go
```diff
@@ -462,6 +465,11 @@ func (c *CriManager) ListPodSandbox(ctx context.Context, r *runtime.ListPodSandb
...
 	}
 
+	sandboxList, err = c.filterInvalidSandboxes(ctx, sandboxList)
+	if err != nil {
+		return nil, fmt.Errorf("failed to filter invalid sandboxes: %v", err)
+	}
+
 	sandboxes := make([]*runtime.PodSandbox, 0, len(sandboxList))
 	for _, s := range sandboxList {
 		sandbox, err := toCriSandbox(s)
```
Starnop: How about put the outermost loop of `filterInvalidSandboxes` here, WDYT?

Expect that, LGTM.

YaoZengzeng: Hmm... It's a code organization issue, I want to encapsulate all the logic into this function.

Now, I think it's OK.

As the code evolve, maybe it will be better to put the loop outside :)

PR #2092, opened 2018-08-14
cgroup: enable memory swappiness equal to -1

Ace-Tang:

enable memory-swappiness can be set to -1, -1 means do nothing.

add！

### DIFF
--- a/apis/opts/memory_swappiness.go
+++ b/apis/opts/memory_swappiness.go
@@ -6,7 +6,7 @@ import "fmt"
 
 // ValidateMemorySwappiness verifies the correctness of memory-swappiness.
 func ValidateMemorySwappiness(memorySwappiness int64) error {
-	if memorySwappiness < 0 || memorySwappiness > 100 {
+	if memorySwappiness != -1 && (memorySwappiness < 0 || memorySwappiness > 100) {
 		return fmt.Errorf("invalid memory swappiness: %d (its range is 0-100)", memorySwappiness)
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


apis/opts/memory_swappiness.go
```diff
@@ -6,7 +6,7 @@ import "fmt"
 
 // ValidateMemorySwappiness verifies the correctness of memory-swappiness.
 func ValidateMemorySwappiness(memorySwappiness int64) error {
-	if memorySwappiness < 0 || memorySwappiness > 100 {
+	if memorySwappiness != -1 && (memorySwappiness < 0 || memorySwappiness > 100) {
```
HusterWan: how about `if  (memorySwappiness < -1 || memorySwappiness > 100)`

daemon/mgr/container_validation.go
```diff
@@ -100,7 +100,7 @@ func validateResource(r *types.Resources, update bool) ([]string, error) {
 			warnings = append(warnings, MemorySwappinessWarn)
 			r.MemorySwappiness = nil
 		}
-		if r.MemorySwappiness != nil && (*r.MemorySwappiness < 0 || *r.MemorySwappiness > 100) {
+		if r.MemorySwappiness != nil && *r.MemorySwappiness != -1 && (*r.MemorySwappiness < 0 || *r.MemorySwappiness > 100) {
```
HusterWan: how about `if r.MemorySwappiness != nil  (*r.MemorySwappiness < -1 || *r.MemorySwappiness > 100)`

apis/opts/memory_swappiness.go
Ace-Tang: Do not do this, [0, 100] is a valid value, -1 is the value we enable to follow with moby.

HusterWan: OK, You are the boss :)

HusterWan:
LGTM

PR #2109, opened 2018-08-17
bugfix: return error when remove/stop containers of sandbox failed

YaoZengzeng:

When stop/remove the sandbox, we should stop/remove the containers of the sandbox as well.

However if we failed to stop/remove the containers, we should report the error to kubelet.

Otherwise if the sandbox is removed successfully and the containers are failed to be removed,

the containers will be left and never be cleaned.


YaoZengzeng:
@fuweid @Starnop PTAL

PR #2119, opened 2018-08-20
bugfix: fix that list volumes sets hostname as default label filter

### LINKED ISSUE #2118: [bug] Don't need to set hostname as default label when list volume
### Ⅰ. Issue Description
List volume sets hostname as default label but create volume did not.

### Ⅱ. Describe what happened
When lists volume, it sets hostname as default label. But it did't set the default one when create volume:
```
// Create is used to create volume.
func (vm *VolumeManager) Create(ctx context.Context, name, driver string, options, labels map[string]string) (*types.Volume, error) {
	if driver == "" {
		driver = types.DefaultBackend
	}

	id := types.VolumeID{
		Name:      name,
		Driver:    driver,
		Options:   map[string]string{},
		Selectors: map[string]string{},
		Labels:    map[string]string{},
	}

	if labels != nil {
		id.Labels = labels
	}

	if options != nil {
		id.Options = options
	}

	v, err := vm.core.CreateVolume(id)
	if err != nil {
		return nil, err
	}

	vm.LogVolumeEvent(ctx, name, "create", map[string]string{"driver": driver})

	return v, nil
}

// List returns all volumes on this host.
func (vm *VolumeManager) List(ctx context.Context, labels map[string]string) ([]*types.Volume, error) {
	if _, ok := labels["hostname"]; !ok {
		hostname, err := os.Hostname()
		if err != nil {
			return nil, err
		}

		labels["hostname"] = hostname
	}

	return vm.core.ListVolumes(labels)
}
```
It worked fine before because the list did not implement the filter function:
```
// ListVolumes return all volumes.
// Param 'labels' use to filter the volumes, only return those you want.
func (c *Core) ListVolumes(labels map[string]string) ([]*types.Volume, error) {
	var retVolumes = make([]*types.Volume, 0)

	// list local meta store.
	metaList, err := c.store.List()
	if err != nil {
		return nil, err
	}

	// scan all drivers.
	logrus.Debugf("probing all drivers for listing volume")
	drivers, err := driver.GetAll()
	if err != nil {
		return nil, err
	}

	ctx := driver.Contexts()

	var realVolumes = map[string]*types.Volume{}
	var volumeDrivers = map[string]driver.Driver{}

	for _, dv := range drivers {
		volumeDrivers[dv.Name(ctx)] = dv

		d, ok := dv.(driver.Lister)
		if !ok {
			// not Lister, ignore it.
			continue
		}
		vList, err := d.List(ctx)
		if err != nil {
			logrus.Warnf("volume driver %s list error: %v", dv.Name(ctx), err)
			continue
		}

		for _, v := range vList {
			realVolumes[v.Name] = v
		}
	}

	for name, obj := range metaList {
		v, ok := obj.(*types.Volume)
		if !ok {
			continue
		}

		d, ok := volumeDrivers[v.Spec.Backend]
		if !ok {
			// driver not exist, ignore it
			continue
		}

		// the local driver and tmpfs driver
		if d.StoreMode(ctx).IsLocal() {
			retVolumes = append(retVolumes, v)
			continue
		}

		rv, ok := realVolumes[name]
		if !ok {
			// real volume not exist, ignore it
			continue
		}
		v.Status.MountPoint = rv.Status.MountPoint

		delete(realVolumes, name)

		retVolumes = append(retVolumes, v)
	}

	for _, v := range realVolumes {
		// found new volumes, store the meta
		logrus.Warningf("found new volume %s", v.Name)
		c.store.Put(v)

		retVolumes = append(retVolumes, v)

	}

	return retVolumes, nil
}
```

Also reference #1817, #2083.

List volume doesn't need to set hostname as default label.

1. Create a volume without labels.
2. List volumes, we can find the default filter label is `hostname`.

It worked fine before because the list did not implement the filter function. 

### Ⅵ. Environment:

- pouch version (use `pouch version`): 1.0.0-rc1
- OS (e.g. from /etc/os-release): CentOS7
- Kernel (e.g. `uname -a`): 3.10.0-327.18.2.el7.x86_64
- Install tools: source code
- Others:

xiaoxubeii:
Remove default label filter when list volumes.

fixes #2118.

Yes.

### DIFF
--- a/daemon/mgr/volume.go
+++ b/daemon/mgr/volume.go
@@ -2,7 +2,6 @@ package mgr
 
 import (
 	"context"
-	"os"
 	"strings"
 
 	"github.com/alibaba/pouch/daemon/events"
@@ -109,15 +108,6 @@ func (vm *VolumeManager) Get(ctx context.Context, name string) (*types.Volume, e
 
 // List returns all volumes on this host.
 func (vm *VolumeManager) List(ctx context.Context, labels map[string]string) ([]*types.Volume, error) {
-	if _, ok := labels["hostname"]; !ok {
-		hostname, err := os.Hostname()
-		if err != nil {
-			return nil, err
-		}
-
-		labels["hostname"] = hostname
-	}
-
 	return vm.core.ListVolumes(labels)
 }
 


rudyfly:
LGTM

PR #2141, opened 2018-08-22
bugfix: fix runtime script may be coverd

Ace-Tang:
if condition should change location, or runtime script path
will be coverd by runtime path.


CodeJuan:
LGTM

PR #2144, opened 2018-08-22
test: add test for client utils

Ace-Tang:
test: add test for client utils

### DIFF
--- a/client/utils_test.go
+++ b/client/utils_test.go
@@ -1,0 +1,39 @@
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


client/utils.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 func decodeBody(obj interface{}, body io.Reader) error {
+	if body == nil {
+		return fmt.Errorf("failed to decode a null body")
```
HusterWan: how about `can not decode a null body` ?

fuweid:
synced with @Ace-Tang , we don't need this pr.

PR #2152, opened 2018-08-23
bugfix: avoid data race in set exec exit status

Ace-Tang:
there are two goroutines response for changed exec config information,
if exec process got error, data race appears, exec exit hook runs in a
goroutine, it modify exec exit status, StartExec also modify exec exit
status in another goroutine since error happend. Fix it by only one
place can modify exec exit status.

### DIFF
--- a/ctrd/container.go
+++ b/ctrd/container.go
@@ -122,14 +122,13 @@ func (c *Client) execContainer(ctx context.Context, process *Process) error {
 			exitTime: status.ExitTime(),
 		}
 
-		if err := <-fail; err != nil {
-			msg.err = err
-		}
-
-		for _, hook := range c.hooks {
-			if err := hook(process.ExecID, msg); err != nil {
-				logrus.Errorf("failed to execute the exec exit hooks: %v", err)
-				break
+		// run hook if not got fail here
+		if err := <-fail; err == nil {
+			for _, hook := range c.hooks {
+				if err := hook(process.ExecID, msg); err != nil {
+					logrus.Errorf("failed to execute the exec exit hooks: %v", err)
+					break
+				}
 			}
 		}
 
--- a/daemon/mgr/container_exec.go
+++ b/daemon/mgr/container_exec.go
@@ -77,7 +77,6 @@ func (mgr *ContainerManager) StartExec(ctx context.Context, execid string, attac
 			exitCode := 126
 			execConfig.ExitCode = int64(exitCode)
 		}
-		mgr.ExecProcesses.Put(execid, execConfig)
 	}()
 
 	if attach != nil {
--- a/test/cli_exec_test.go
+++ b/test/cli_exec_test.go
@@ -144,12 +144,23 @@ func (suite *PouchExecSuite) TestExecExitCode(c *check.C) {
 	command.PouchRun("exec", name, "sh", "-c", "exit 0").Assert(c, icmd.Success)
 }
 
-// TestExecFail test exec fail should not hang.
+// TestExecFail test exec fail should not hang, and test failed exec exit code should not be zero.
 func (suite *PouchExecSuite) TestExecFail(c *check.C) {
 	name := "TestExecFail"
 	res := command.PouchRun("run", "-d", "--name", name, "-u", name, busyboxImage, "top")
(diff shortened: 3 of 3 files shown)


Ace-Tang:
The result may error depend on which goroutine run last
```
#./pouch exec 30f44535e526947d9e1cb41974f4fb944c596c385938d353002f0e44646ea7ea aaaaa
failed to exec process: OCI runtime exec failed: exec failed: container_linux.go:265: starting container process caused "exec: \"aaaaa\": executable file not found in $PATH": unknown

#echo $?
0
```
#./pouch exec 30f44535e526947d9e1cb41974f4fb944c596c385938d353002f0e44646ea7ea dsad
failed to exec process: OCI runtime exec failed: exec failed: container_linux.go:265: starting container process caused "exec: \"dsad\": executable file not found in $PATH": unknown

#echo $?
126

```

File: ctrd/container.go [hunk 1]
```diff
@@ -122,14 +122,13 @@ func (c *Client) execContainer(ctx context.Context, process *Process) error {
...
-		if err := <-fail; err != nil {
-			msg.err = err
-		}
-
-		for _, hook := range c.hooks {
-			if err := hook(process.ExecID, msg); err != nil {
-				logrus.Errorf("failed to execute the exec exit hooks: %v", err)
-				break
+		// run hook if not got fail here
+		if err := <-fail; err == nil {
```
yyb196: can we get the real exit code?

Ace-Tang: When fails here, we got a empty(meaningless) exit status, that means exit code is zero.

test/cli_exec_test.go
```diff
@@ -150,6 +150,11 @@ func (suite *PouchExecSuite) TestExecFail(c *check.C) {
 	res := command.PouchRun("run", "-d", "--name", name, "-u", name, busyboxImage, "top")
 	defer DelContainerForceMultyTime(c, name)
 	c.Assert(res.Stderr(), check.NotNil)
+
+	// test a 'executable file not found' fail should get exit code 126
+	name = "TestExecFailCode"
+	command.PouchRun("run", "-d", "--name", name, busyboxImage, "top").Assert(c, icmd.Success)
```
fuweid: could you update the case description, since there are two cases here?

Ace-Tang:
fail with
```
" exited with 0.
$ bash <(curl -s https://codecov.io/bash) -cF ${TEST_SUITE} -y .codecov.yml
/dev/fd/63: line 1: html: No such file or directory
/dev/fd/63: line 2: syntax error near unexpected token `<'
```

test/cli_exec_test.go
```diff
@@ -144,12 +144,23 @@ func (suite *PouchExecSuite) TestExecExitCode(c *check.C) {
...
 
-// TestExecFail test exec fail should not hang.
+// TestExecFail test exec fail should not hang, and test failed exec exit code should not be zero.
 func (suite *PouchExecSuite) TestExecFail(c *check.C) {
 	name := "TestExecFail"
 	res := command.PouchRun("run", "-d", "--name", name, "-u", name, busyboxImage, "top")
 	defer DelContainerForceMultyTime(c, name)
 	c.Assert(res.Stderr(), check.NotNil)
+
+	// test a 'executable file not found' fail should get exit code 126.
```
fuweid: could we split it into two cases?

daemon/mgr/container_exec.go
```diff
@@ -77,7 +77,6 @@ func (mgr *ContainerManager) StartExec(ctx context.Context, execid string, attac
 			exitCode := 126
 			execConfig.ExitCode = int64(exitCode)
 		}
-		mgr.ExecProcesses.Put(execid, execConfig)
```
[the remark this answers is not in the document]
fuweid: For this part, I want to raise proposal that we should deep copy data from manager data store to avoid the data race. But it will introduce the complicated.

test/cli_exec_test.go
Ace-Tang: Just keep this ?

Ace-Tang:
@fuweid, deep copy may need, but for this case, we should keep the data race from the logic, deep copy won't avoid this.

yyb196:
LGTM

PR #2193, opened 2018-09-03
bugfix: set pid of stopped container to 0

HusterWan:

We set pid of stopped container to `-1` before, it's ok that we do not check value of pid but the status of container when we want to check whether the container is stopped or not.

But now, `pouchd` supported `CRI`, the k8s will check the value of pid when the container stopped and the pid is `0` when container stopped.

So, in order to stay with moby and kubernetes, we also should change the pid to `0` when the container is stopped.

none

added `TestStopPidValue` test case

```
root@osboxes:test (zr/pid-fix *) -> pouch run -d -t  centos top
d42b8e1c8c45dbdbf53f1fb2432ba6a0c2d730007601c8d47f2bdf94508f8c58
root@osboxes:test (zr/pid-fix) -> pouch ps
Name     ID       Status         Created         Image                                           Runtime
d42b8e   d42b8e   Up 3 seconds   4 seconds ago   registry.hub.docker.com/library/centos:latest   runc
root@osboxes:test (zr/pid-fix) -> pouch stop -t 3 d42b8e
d42b8e
root@osboxes:test (zr/pid-fix) -> pouch inspect -f '{{.State.Pid}}' d42b8e
0
```

None

### DIFF
--- a/test/cli_stop_test.go
+++ b/test/cli_stop_test.go
@@ -126,3 +126,36 @@ func (suite *PouchStopSuite) TestStopMultiContainers(c *check.C) {
 	c.Assert(string(result[0].State.Status), check.Equals, "stopped")
 
 }
+
+// TestStopPidValue ensure stopped container's pid is 0
+func (suite *PouchStopSuite) TestStopPidValue(c *check.C) {
+	name := "test-stop-pid-value"
+
+	command.PouchRun("run", "-d", "--name", name, busyboxImage, "top").Assert(c, icmd.Success)
+	defer DelContainerForceMultyTime(c, name)
+
+	// test stop a created container
+	command.PouchRun("stop", name).Assert(c, icmd.Success)
+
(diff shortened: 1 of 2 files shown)


test/cli_stop_test.go
```diff
@@ -126,3 +126,21 @@ func (suite *PouchStopSuite) TestStopMultiContainers(c *check.C) {
 	c.Assert(string(result[0].State.Status), check.Equals, "stopped")
 
 }
+
+// TestStopPidValue ensure stopped container's pid is 0
+func (suite *PouchStopSuite) TestStopPidValue(c *check.C) {
```
fuweid: please add one case for the process exits by itself.

HusterWan: make sense

rudyfly:
LGTM

allencloud:
You have filled up the beautiful pull request description. @HusterWan

PR #2200, opened 2018-09-05
bugfix: sort mounts according to the depth of destination path

YaoZengzeng:

sortMounts sorts an array of mounts in lexicographic order.

This ensure that when mounting, the mounts don't shadow other mounts. 

For example, if mounting /etc and /etc/resolv.conf, /etc/resolv.conf must not be mounted first.

ref: https://github.com/containerd/containerd/issues/2596

Added


fuweid:
wait for CI

PR #2205, opened 2018-09-05
fix: simplify logic of update env file, and fix bug

Ace-Tang:
since updateContainerEnv logic is too complicated,
content in /etc/profile.d/pouchenv.sh should same with config env,
cover env file with merged env.
fix check same value error:
'msg="env exists and the value is not same, key=spas_accessKey, old
value="4f032bc4b0db47db9baa95fcd4c7b3ad", new value=4f032bc4b0db47db9baa95fcd4c7b3ad"'

fix failed to update env file with old logic.

add rich container update env ci test.


daemon/mgr/container_envs.go
```diff
@@ -24,110 +20,40 @@ import (
...
-		return nil
-	}
-	if _, err := os.Stat(envShPath); err != nil {
-		logrus.Warnf("failed to stat container's env file /etc/profile.d/pouchenv.sh: %v", err)
+func updateContainerEnv(env []string, baseFs string) error {
+	// if target env directory is not exist, return
+	// generally speaking, we should check env file exist to decide
+	// whether to update env file, but in case of missing something,
+	// we check directory existence here to make a decision.
+	if _, err := os.Stat(filepath.Join(baseFs, filepath.Dir(pouchEnvFile))); err != nil {
```
HusterWan: why we not check the file existence, but check the dir?

daemon/mgr/container_envs.go
```diff
@@ -24,110 +20,40 @@ import (
...
+			s = strings.Replace(s, "\"", "\\\"", -1)
+			s = strings.Replace(s, "$", "\\$", -1)
+			if parts[0] == "PATH" {
+				s = parts[1] + ":$PATH"
+			}
+			str += fmt.Sprintf("export %s=\"%s\"\n", parts[0], s)
 		}
 	}
-	return localEnv
+	ioutil.WriteFile(filepath.Join(baseFs, pouchEnvFile), []byte(str), 0755)
```
HusterWan: OK, i got you, we will just write the env to the file, so we just need check the existence of Dir

daemon/mgr/container.go
```diff
@@ -999,16 +1006,12 @@ func (mgr *ContainerManager) Update(ctx context.Context, name string, config *ty
...
-	if err != nil {
-		return err
-	}
-	c.Config.Env = newEnvSlice
-
 	if mgr.containerPlugin != nil {
-		if err = mgr.containerPlugin.PostUpdate(c.BaseFS, c.Config.Env); err != nil {
-			return err
+		// post update env file
+		if len(config.Env) > 0 {
```
HusterWan: how about `if len(config.Env) > 0 && err := mgr.containerPlugin.PostUpdate(c.BaseFS, c.Config.Env); err != nil` ?

Ace-Tang: Too long, what do you think?

Ace-Tang: Look like conbine `mgr.containerPlugin != nil && len(config.Env) > 0` better?

HusterWan: can not agree no more

Ace-Tang: Updated.

HusterWan:
lgtm

daemon/mgr/container.go
```diff
@@ -989,6 +989,13 @@ func (mgr *ContainerManager) Update(ctx context.Context, name string, config *ty
 	}
 	c.Unlock()
 
+	// Update Env
```
HusterWan: update container parameters, should we add `Lock` here?

Ace-Tang: This should be disscuss in another pr, we are not determine to modify origin logic here, or you can raise a pr if  you think if lock should be added.

PR #2224, opened 2018-09-10
fix: fix cri exec hang

Ace-Tang:
when exec fails, io should get from mgr.IOs.


Ace-Tang:
If exec process get run, io should be closed in the function , still have no iead why it can't close backend after function.  If io closed after the function, log followed will miss.
```
close containerio backend: hijack, id: c7b8106b3420b1b27e3dc3996a151ce0e4db2bec1d2e60730bb479650ce31cae
```

ctrd/container.go
```diff
@@ -122,13 +122,18 @@ func (c *Client) execContainer(ctx context.Context, process *Process) error {
...
-		if err := <-fail; err == nil {
-			for _, hook := range c.hooks {
-				if err := hook(process.ExecID, msg); err != nil {
-					logrus.Errorf("failed to execute the exec exit hooks: %v", err)
-					break
-				}
+		if err := <-fail; err != nil {
+			msg.err = err
+			// exit code should not be zero when exec get failed.
+			if msg.exitCode == 0 {
```
HusterWan: whether it is possible `exitCode` is not zero when `exec` got error?

Ace-Tang: It possible, like `exec $id aaaaa` .

ctrd/container.go
```diff
@@ -122,13 +122,18 @@ func (c *Client) execContainer(ctx context.Context, process *Process) error {
...
+				msg.exitCode = 126
+			}
+		}
+		// XXX: if exec process get run, io should be closed in this function,
+		for _, hook := range c.hooks {
```
HusterWan: Now, no matter run `exec` got an error or not, we always call the exit hook.

HusterWan:
lgtm

ctrd/container.go
Ace-Tang: yes, when exec start , no matter what result, let hook do the clean. Before exec start, if got error, let pouch daemon level do the clean.

Ace-Tang:
@allencloud , merge this before #2226 pass ci.

Starnop:
@Ace-Tang  All tests has been passed. ^_^

Ace-Tang:
@Starnop , thanks for test ! I think we can merge #2226 now.

PR #2238, opened 2018-09-13
bugfix: fix lock competition between Stats and Start interface

HusterWan:

We use `Stats` interface to get running container's resource metrics, and this api can be called at any time of the container's life cycle.

If we called the `Stats` interface when the container is starting, there may caused `Stats`  or `Start` failed because of lock competition.

The code that get the lock:
```
     if !c.lock.Trylock(id) {
       }
       defer c.lock.Unlock(id)
  
```

The log that get lock failed:

```
time="2018-09-12T16:57:10.49956582+08:00" level=error msg="failed to decode metrics of container "b36804c94a701cfb5f8e3c27c6e9eed096e8179f6352ec465df156efb01b28c8": failed to get stats of container "b36804c94a701cfb5f8e3c27c6e9eed096e8179
f6352ec465df156efb01b28c8": lock failed"
```

```
500 Internal Server Error: {"message":"failed to create container(${CNRID}) on containerd: lock failed"}
```

So, in order to avoid the lock competition, we should just get the `Metrics` just when the container is running

none

This situation is very hard to write test case.

none

none

### DIFF
--- a/daemon/mgr/container_stats.go
+++ b/daemon/mgr/container_stats.go
@@ -59,10 +59,15 @@ func (mgr *ContainerManager) StreamStats(ctx context.Context, name string, confi
 			if err != nil {
 				return err
 			}
-			containerStat := toContainerStats(metrics.Timestamp, stats)
 
-			if err := enc.Encode(containerStat); err != nil {
-				return err
+			// metrics may be nil if the container is not running,
+			// so just ignore it and try get the metrics next time.
+			if metrics != nil {
+				containerStat := toContainerStats(metrics.Timestamp, stats)
+
+				if err := enc.Encode(containerStat); err != nil {
+					return err
+				}
 			}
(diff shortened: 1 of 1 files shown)


daemon/mgr/container_stats.go
```diff
@@ -81,6 +81,11 @@ func (mgr *ContainerManager) Stats(ctx context.Context, name string) (*container
 	ID := c.ID
 	c.Unlock()
 
+	// only get metrics when the container is running
+	if !(c.IsRunning() || c.IsPaused()) {
+		return nil, nil, nil
```
fuweid: `return nil` will panic if the StreamStat `toContainerStats` consumes the result.

HusterWan: make sense, i have fix it.

PR #2240, opened 2018-09-13
bugfix: fix that pouch don't support detach-keys when run "pouch run"

### LINKED ISSUE #2221: [bug] detatch-keys config can't work
### Ⅰ. Issue Description
pouch run --detach-keys can't work

### Ⅱ. Describe what happened
`pouch run -it --detach-keys="ctrl-z" busybox` while attached into container, ctrl-z can't detach from container. But with docker, it could work.

`pouch run -it --detach-keys="ctrl-z" busybox`
when type ctrl-z, detach cloud happen and container is still active.

1.`pouch run -it --detach-keys="ctrl-z" busybox`
2. type ctrl-z
3.

May be related the issue I raise #2218 

### Ⅵ. Environment:

```
GitCommit:       1.0.0-28-gcdf6e8e-dirty
GoVersion:       go1.9.2
KernelVersion:   4.15.0-33-generic
Os:              linux
Version:         1.0.0
APIVersion:      1.24
Arch:            amd64
BuildTime:       2018-09-08T15:48:20+08:00
```
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

xiaoxubeii:
`pouch run` did't support `--detach-keys`

fixes #2221

```
$ pouch run -it --detach-keys="ctrl-x" busybox
```

It seems that pouch lacks a unified detach mechanism to capture custom detach keys. The following commands will be affected:
- pouch run 
- pouch exec
- pouch start

I will submit a series of pr to solve them later.

### DIFF
--- a/cli/run.go
+++ b/cli/run.go
@@ -16,6 +16,9 @@ import (
 var runDescription = "Create a container object in Pouchd, and start the container. " +
 	"This is useful when you just want to use one command to start a container. "
 
+// The default escape key sequence: ctrl-p, ctrl-q
+var defaultEscapeKeys = []byte{16, 17}
+
 // RunCommand use to implement 'run' command, it creates and starts a container.
 type RunCommand struct {
 	baseCommand
@@ -125,12 +128,33 @@ func (rc *RunCommand) runRun(args []string) error {
 		}
 		defer conn.Close()
 
+		// TODO wait for https://github.com/alibaba/pouch/pull/2258 to merge
+		// escapeKeys := defaultEscapeKeys
+		// // Wrap the input to detect detach escape sequence.
+		// // Use default escape keys if an invalid sequence is given.
+		// if rc.detachKeys != "" {
+		// 	customEscapeKeys, err := term.ToBytes(rc.detachKeys)
+		// 	if err != nil {
+		// 		logrus.Warnf("invalid detach escape keys, using default: %s", err)
+		// 	} else {
+		// 		escapeKeys = customEscapeKeys
+		// 	}
+		// }
+
+		// in := ioutils.NewReadCloserWrapper(term.NewEscapeProxy(os.Stdin, escapeKeys), os.Stdin.Close)
+
 		go func() {
 			io.Copy(os.Stdout, br)
 			wait <- struct{}{}
 		}()
 		go func() {
 			io.Copy(conn, os.Stdin)
+			// TODO wait for https://github.com/alibaba/pouch/pull/2258 to merge
+			// _, err := io.Copy(conn, in)
+			// if _, ok := err.(term.EscapeError); ok {
+			// 	wait <- struct{}{}
+			// 	return
+			// }
 		}()
 	}
 
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -800,6 +800,12 @@
 			"revisionTime": "2018-06-06T04:43:31Z"
 		},
 		{
+			"checksumSHA1": "E6YGk+mLX+kID6fxKWNN6gbH6eg=",
+			"path": "github.com/docker/docker/pkg/term",
+			"revision": "fd2f2a919e392b96de74795ae9af2dc5e510bc4c",
+			"revisionTime": "2018-06-06T04:43:31Z"
+		},
+		{
 			"checksumSHA1": "1IPGX6/BnX7QN4DjbBk0UafTB2U=",
 			"path": "github.com/docker/go-connections/nat",
 			"revision": "7395e3f8aa162843a74ed6d48e79627d9792ac55",
--- a/vendor/github.com/docker/docker/pkg/term/ascii.go
+++ b/vendor/github.com/docker/docker/pkg/term/ascii.go
@@ -1,0 +1,66 @@
+package term // import "github.com/docker/docker/pkg/term"
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ASCII list the possible supported ASCII key sequence
+var ASCII = []string{
+	"ctrl-@",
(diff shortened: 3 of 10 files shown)


cli/run.go
```diff
@@ -125,12 +128,28 @@ func (rc *RunCommand) runRun(args []string) error {
 		}
 		defer conn.Close()
 
+		var in io.Reader = os.Stdin
+		// Wrap the input to detect detach escape sequence.
+		// Use default escape keys if an invalid sequence is given.
```
ZYecho: where's the default escape keys?

pkg/term/proxy.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// Looks like we've got an escape key, but we need to match again on the next
+	// read.
+	// Store the current escape key we found so we can look for the next one on
+	// the next read.
+	// Since this is an escape key, make sure we don't let the caller read it
+	// If later on we find that this is not the escape sequence, we'll add the
+	// keys back
+	r.escapeKeyPos++
```
ZYecho: Why the code execute here imply got an escape key?  if the read bytes get "..... detach-bytes" which detach bytes are concated at last, why escapePos move?

pkg/term/ascii.go
```diff
@@ -0,0 +1,66 @@
+package term
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ASCII list the possible supported ASCII key sequence
+var ASCII = []string{
```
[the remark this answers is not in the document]
ZYecho: So if we in the future decide to add more ctrl-1~9 detach-keys support, it will be add nine constants here? I mean It's will be a long list and not easy to review and check.

pkg/term/proxy.go
xiaoxubeii: Because it allows specify multiple keys by `pouch run -it --detach-keys="q,q,q,q,q,q,q"`

> Why the code execute here imply got an escape key? if the read bytes is "..... detach-bytes" for example which detach bytes are concated at last,how r.escapePos move?

Because it allows specify multiple keys by `pouch run -it --detach-keys="q,q,q,q,q,q,q"`, the escape keys would be `[]byte`. We must match the following inputs.

pkg/term/ascii.go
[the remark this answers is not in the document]
ZYecho: LGTM,  looks same as [docker](https://github.com/docker/docker-ce/blob/8a582e89b407639018e067123eb17f61dbfd10ee/components/engine/pkg/term/ascii.go)

pkg/term/proxy.go
ZYecho: LGTM

xiaoxubeii:
@fuweid Can you retest it ? I don't think my code is causing the test to fail.

pkg/term/proxy.go
```diff
@@ -0,0 +1,81 @@
+package term
```
fuweid: could we just use `github.com/docker/docker/pkg/term` package?

cli/run.go
```diff
@@ -125,12 +131,31 @@ func (rc *RunCommand) runRun(args []string) error {
...
+		in = ioutils.NewReadCloserWrapper(term.NewEscapeProxy(os.Stdin, escapeKeys), os.Stdin.Close)
+
 		go func() {
 			io.Copy(os.Stdout, br)
 			wait <- struct{}{}
 		}()
 		go func() {
-			io.Copy(conn, os.Stdin)
+			_, err := io.Copy(conn, in)
+			if _, ok := err.(term.EscapeError); ok {
```
fuweid: @xiaoxubeii could you mind to provide more detail about this kind of error? what condition will trigger this error?

for now, we cannot just send the signal to the channel to quit `run` command because we need to send the close stdin api to containerd. I provide more detail in the PR #2258. In my opinion, keep it. just print the error and return. we can fix it in the following PR. What do you think about this?

vendor/vendor.json
```diff
@@ -799,6 +799,12 @@
 			"revision": "fd2f2a919e392b96de74795ae9af2dc5e510bc4c",
 			"revisionTime": "2018-06-06T04:43:31Z"
 		},
+		{
+			"checksumSHA1": "WjvsjU1rtFjD3S0MmzKi5M08zjc=",
+			"path": "github.com/docker/docker/pkg/term",
+			"revision": "e68fc7a215d7133c34aa18e3b72b4a21fd0c6136",
+			"revisionTime": "2018-08-21T03:16:31Z"
```
fuweid: I think the `pkg/term` should keep consistent with existing moby package here.

cli/run.go
xiaoxubeii: > @xiaoxubeii could you mind to provide more detail about this kind of error? what condition will trigger this error?
> 
> for now, we cannot just send the signal to the channel to quit `run` command because we need to send the close stdin api to containerd. I provide more detail in the PR #2258. In my opinion, keep it. just print the error and return. we can fix it in the following PR. What do you think about this?

It will return `term.EscapeError` when catches a detach keys:
https://github.com/alibaba/pouch/blob/f285f1863903ba4bc0778b45d0d194786ce9e98a/vendor/github.com/docker/docker/pkg/term/proxy.go#L65-L67

OK, I would print the error instead of closing the io, we can fix it later.

xiaoxubeii:
@fuweid Done

xiaoxubeii:
@allencloud @fuweid Kindly ping for review and merge : )

fuweid:
synced with @xiaoxubeii offline. we close it and will reopen it when the refactor work is ready.

PR #2245, opened 2018-09-17
bugfix: add check in case empty detach-keys override

ZYecho:

as title mentioned,  effect pouch start/run command.

None

No need

None

None


ZYecho:
@allencloud I saw the Container struct has field DetachKeys, I think it's useless. Maybe remove it is a good way? In docker, if you can't override the default detach-keys when get a terminal such as run/start, the detach-keys will be the default not same as the former one.

fuweid:
@ZYecho do we have test case here?

ZYecho:
close it temporarily

PR #2254, opened 2018-09-19
cri: add default annotation sandboxid

Ace-Tang:
for vm level runtime, add default annotation
io.kubernetes.cri-o.SandboxID when create container
in sanbox.

As for kata, the default annotation is
```
// CRIContainerTypeKeyList lists all the CRI keys that could define
    // the container type from annotations in the config.json.
    CRIContainerTypeKeyList = []string{criContainerdAnnotations.ContainerType, crioAnnotations.ContainerType, dockershimAnnotations.ContainerTypeLabelKey}

    // CRISandboxNameKeyList lists all the CRI keys that could define
    // the sandbox ID (sandbox ID) from annotations in the config.json.
    CRISandboxNameKeyList = []string{criContainerdAnnotations.SandboxID, crioAnnotations.SandboxID, dockershimAnnotations.SandboxIDLabelKey}
```
I also wonder whether cri mistake `SandboxName` as `SandboxID` as default annotation, @YaoZengzeng @Starnop 
```
specAnnotation[anno.SandboxName] = podSandboxID
```

no.

### DIFF
--- a/cri/annotations/annotations.go
+++ b/cri/annotations/annotations.go
@@ -14,6 +14,9 @@ const (
 	// SandboxName is the sandbox name annotation
 	SandboxName = "io.kubernetes.cri-o.SandboxName"
 
+	// SandboxID is the sandbox id annotation
+	SandboxID = "io.kubernetes.cri-o.SandboxID"
+
 	// KubernetesRuntime is the runtime
 	KubernetesRuntime = "io.kubernetes.runtime"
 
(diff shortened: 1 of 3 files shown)


cri/annotations/annotations.go
```diff
@@ -14,6 +14,9 @@ const (
 	// SandboxName is the sandbox name annotation
 	SandboxName = "io.kubernetes.cri-o.SandboxName"
```
Ace-Tang: Actually, I think this shoud be `SanboxID` , but not `SanboxName`.

Starnop: It will use `SandboxName` as sharedContainer for runv, please refer to https://github.com/hyperhq/runv/blob/master/cli/create.go#L96

Ace-Tang: OK, since kata need `SandboxID`, just append it.

PR #2274, opened 2018-09-25
bugfix:  event api connection may be recycled if keeps idle too long

HusterWan:

If pouchd  start http server with parameters below, http server will recycle the long connection if it keeps idle too long.
 
```
			s := &http.Server{
				Handler:           router,
				ErrorLog:          log.New(stdFilterLogWriter, "", 0),
				ReadTimeout:       time.Minute * 10,
				ReadHeaderTimeout: time.Minute * 10,
				WriteTimeout:      time.Minute * 10,
				IdleTimeout:       time.Minute * 10,
			}
```

So, i add a ticker in `event` api too keep the connection alive.

none

none

none

none


apis/server/system_bridge.go
```diff
@@ -123,11 +121,17 @@ func (s *Server) events(ctx context.Context, rw http.ResponseWriter, req *http.R
 		return nil
 	}
 
+	// Notes(ziren): The pouchd http server set some http timeout parameters (see pouch/apis/server/server.go),
+	// that may cause long idle connections been recycled. So, creating a tick here to keep the
+	// event api alive. set the ticker to 8s because of the IdleTimeout of http server is 10s.
+	tickChan := time.NewTicker(time.Second * 10).C
```
fuweid: I think we should use `time.NewTicker` and `defer tick.Stop()` here, because we don't want to leave it active in the timer heap.

apis/server/system_bridge.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/alibaba/pouch/apis/types"
 	"github.com/alibaba/pouch/pkg/httputils"
 	"github.com/alibaba/pouch/pkg/utils"
+	"github.com/sirupsen/logrus"
```
fuweid: move it into next group

HusterWan: good catch

apis/server/system_bridge.go
HusterWan: I am afraid we may not need the `defer` because i checked the example of `NewTicker` when we should add `defer tick.Stop()` when using `NewTicker`, but not need the `Stop` when using `NewTicker().C`

HusterWan: OK, when i read the test cases in `time` package, i realize i am wrong. we should call `defer tick.Stop()` when creat a ticker.

ctrd/client.go
```diff
@@ -384,15 +384,15 @@ func (c *Client) collectContainerdEvents() {
 		}
 
 		switch e.Topic {
-		case ContainersDeleteEventTopic:
-			cDelEvent, ok := out.(*eventstypes.ContainerDelete)
+		case TaskExitEventTopic:
```
HusterWan: Need know the exit code when a container exited, so we can get the task exit code from containerd `task/exit` event

HusterWan:
events api connection broken because of the `WriteTimeout` parameter of http server, when i delete the `WriteTimeout`, the connection never die.

Refer: https://github.com/golang/go/issues/24461

PR #2310, opened 2018-10-15
bugfix: remove set quota for container meta directory

zhuangqh:

Remove set quota for container's metadata directory:
`$graph/containers/$id/`

fix the method to get the device mount point, just get the shortest path
of mount point, such as:
```
/dev/sdb1 /home/pouch ext4 rw,relatime,prjquota,data=ordered 0 0
/dev/sdb1 /home/pouch/overlay ext4 rw,relatime,prjquota,data=ordered 0 0
```
We will choose `/home/pouch` as the mountpoint instead of
`/home/pouch/overlay`.

remove a test case: TestRunWithDiskQuotaForLog

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -1168,17 +1168,9 @@ func (mgr *ContainerManager) updateContainerDiskQuota(ctx context.Context, c *Co
 			}
 		}()
 	}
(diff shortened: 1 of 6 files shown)


rudyfly:
LGTM

PR #2316, opened 2018-10-16
refactor: modify the definition of ContainerPlugin interface

zhuangqh:
Signed-off-by: Wei Fu fhfuwei@163.com

modify the definition of ContainerPlugin interface: PreCreate

Type `*types.ContainerCreateConfig` would be more generic, while `io.ReadCloser` read only stream data.

The CRI need to support the rich-mode container so that the create_hook_plugin should be handled in container_mgr.

redesign the interface

### DIFF
--- a/apis/plugins/ContainerPlugin.go
+++ b/apis/plugins/ContainerPlugin.go
@@ -1,12 +1,16 @@
 package plugins
 
-import "io"
+import (
+	"io"
+
+	"github.com/alibaba/pouch/apis/types"
+)
 
 // ContainerPlugin defines places where a plugin will be triggered in container lifecycle
 type ContainerPlugin interface {
 	// PreCreate defines plugin point where receives a container create request, in this plugin point user
 	// could change the container create body passed-in by http request body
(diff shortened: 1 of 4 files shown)


apis/plugins/ContainerPlugin.go
```diff
@@ -1,12 +1,16 @@
...
+
+	"github.com/alibaba/pouch/apis/types"
+)
 
 // ContainerPlugin defines places where a plugin will be triggered in container lifecycle
 type ContainerPlugin interface {
 	// PreCreate defines plugin point where receives a container create request, in this plugin point user
 	// could change the container create body passed-in by http request body
-	PreCreate(io.ReadCloser) (io.ReadCloser, error)
+	PreCreate(*types.ContainerCreateConfig) error
```
xiaoxubeii: Interface signature and return value are all different from origin, some files missing?

zhuangqh: @xiaoxubeii You are right...

PR #2338, opened 2018-10-21
bugfix: can't allow attach in non-tty client when container in ttyMode

ZYecho:

As the title told. The current implementation have two problems:
1. setRawMode func didn't check container if in ttyMode or not.
2. if you start(with open stdin) a container from a non-tty client with container in tty mode, pouch didn't work and not return any error. this pr fix it by add a checktty func to throw error.

None.

add an integration test TestStartInTTY.

None.

run cli also have same problems, will fix in another pr.

### DIFF
--- a/cli/start.go
+++ b/cli/start.go
@@ -60,7 +60,6 @@ func (s *StartCommand) addFlags() {
 func (s *StartCommand) runStart(args []string) error {
 	ctx := context.Background()
 	apiClient := s.cli.Client()
-
 	// attach to io.
 	if s.attach || s.stdin {
 		var wait chan struct{}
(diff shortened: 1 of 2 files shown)


cli/start.go
```diff
@@ -69,17 +68,24 @@ func (s *StartCommand) runStart(args []string) error {
...
 		if err != nil {
-			return fmt.Errorf("failed to set raw mode")
+			return err
 		}
-		defer func() {
-			if err := restoreMode(in, out); err != nil {
-				fmt.Fprintf(os.Stderr, "failed to restore term mode")
+
+		if c.Config.Tty && terminal.IsTerminal(int(os.Stdout.Fd())) {
+			in, out, err := setRawMode(s.stdin, false)
```
fuweid: I think we should return error here if the input/output device is not terminal. How do you think?

Ace-Tang:
@ZYecho ,  the title not tell where is wrong, wdyt?

cli/start.go
ZYecho: ok, update.

ZYecho:
> @ZYecho , the title not tell where is wrong, wdyt?

Yeah, I update the description of the pr, does fix it?

fuweid:
@ZYecho could you mind to help us to fix the same issue in the exec command? 😄 thanks

PR #2339, opened 2018-10-21
Bump Go versions

HaraldNordgren:


HaraldNordgren:
I'm seeing now that the CLA has lots of Chinese text which I cannot read and I'm not too comfortable about signing an agreement I don't understand. Not sure if there is some way around that, otherwise maybe it's better we close the PR and let some other contributor handle this code change? 🙂

fuweid:
@HaraldNordgren I close it because of duplicated and please let me know if you have any question.

PR #2351, opened 2018-10-24
bugfix: can't allow exec attach ttyMode container from a non-tty client

ZYecho:

fix the same issue with pr #2338 

fix the same issue with pr #2338 

add TestExecWithTty testcase.

None.

None.


test/cli_exec_test.go
```diff
@@ -242,3 +243,14 @@ func (suite *PouchExecSuite) TestExecWithWorkingDir(c *check.C) {
 		c.Fatalf("failed to run exec with specified working directory: %s, but got %s", dir, res.Stdout())
 	}
 }
+
+// TestExecWithTty tests running container with -tty flag and attach stdin in a non-tty client.
+func (suite *PouchExecSuite) TestExecWithTty(c *check.C) {
+	name := "TestExecWithTty"
+	res := command.PouchRun("run", "-d", "--name", name, busyboxImage, "sleep", "100000")
```
fuweid: it's just nit. could we use one line to check `command.PouchRun("run", "-d", "--name", name, busyboxImage, "sleep", "100000").Assert(c, icmd.Success)` ？ thanks.

ZYecho: Done.

cli/exec.go
```diff
@@ -77,6 +77,10 @@ func (e *ExecCommand) runExec(args []string) error {
 		Env:          e.Envs,
 	}
 
+	if err := checkTty(createExecConfig.AttachStdin, createExecConfig.Tty, os.Stdout.Fd()); err != nil {
```
fuweid: I think it should be os.Stdin.Fd() here.

ZYecho: updated.

PR #2363, opened 2018-10-26
feature: support multiple containerd-shim

Ace-Tang:
different runtime can have a specified containerd-shim, especially for
vm runtime like kata/runv.

### what the pr want to do

First we see how pouch run a container, `pouch` -> `containerd` -> `containerd-shim` -> `runtime`. The PR specify the shim can be choosed for the runtime, this is useful for vm runtime like kata/runv, since vm containers have many different with runc container cause they running in a vm machine. 

containerd 1.2 have support a shim v2 API(https://github.com/containerd/containerd/pull/2434), this allowed a runtime to specify itself shim, they can do more custom then.

### what the PR status

The PR is compele finished in function, we pass a shim interface to containerd to choose the shim, this is supported by containerd 1.0.3, so it can be merged now, no need to wait for containerd 1.2. And all process in create container in pouch is keeping same.

### TODO list

1.  kata implement about containerd-shim-v2 is waiting for merged (https://github.com/kata-containers/runtime/pull/572), I have testd it with containerd 1.2, it works good. Kata test may not easy to add in pouch , but we can add some v2 test after we support containerd 1.2.

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -2004,6 +2004,11 @@ definitions:
         items:
           type: "string"
         example: ["--debug", "--systemd-cgroup=false"]
+      shim:
+        description: |
+          shim name, pass to containerd to choose which shim is used for container.
+        type: "string"
+
 
   Commit:
     description: |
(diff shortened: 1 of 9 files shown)


Ace-Tang:
@fuweid , how about this pr, wait so long then I rebase twice?

Ace-Tang:
Hold this pr, the API interface  should be disscussed more

PR #2375, opened 2018-10-28
[RFC] refactor: redesign container io in pouch

fuweid:

Redesign container io in pouch. but Why?

The original design always uses `ringbuffer` to cache the data which causes streams lost data sometime. It makes the regression test unstable. 

It also makes caller hard to touch data from containerd-shim. That is why we still miss the ReopenLog in CRI. I think we should fix that.

#### design overview

pouchcontainer io

Each container/process will hold streams struct to handle the io stream. The struct is like:

```
// Stream is used to handle container IO.
type Stream struct {
        sync.WaitGroup
        stdin          io.ReadCloser
        stdinPipe      io.WriteCloser
        stdout, stderr *multiWriter
}
```

In the stream struct, both the `stdout` and `stderr` are the multiple writer's hub. It means that the data from stdout/stderr of process will have more copies. For example, when user use `run` to start container, the container's output data will have two copies. One is for the terminal side, the other one is for container's log, such as json-file by default.

But for the `stdin`, we don't need the hub like `stdout/stderr`, because the input of process is only one. Any attach requests can send the data to the process in container. So we exports the `stdinPipe` to allow different attach requests to touch the same input channel.

Based on this, the stream acts like bridge to connect attach request and stream of process in container. If the user wants to attach running container, just adds hijack/stream into the output hub and attach it to the `stdinPipe`. The stream can provides more flexible than before.

Before this change, the CRI struggles to integrating with Container Mgr. Why? because CRI component doesn't know the stream has been finished. Always use dead loop to check the `Exec` status. It's not good patten.

#### why add logdriver/crilog into the writer hub?

pouchcontainer io 1

Both logdriver and crilog don't act like `io.Writer`. It always add meta data for the raw data. Adding the partial label is one of examples for the acting way of logdriver. In order to avoid to add duplicate code for each driver, we use `LogCopier` to handle this.

There is a little bit different between `crilog` and `logdriver`. The `crilog` has `ReopenLog` request. It means that it will add new crilog and close previous one. 

In current design, we might have duplicate data in the crilog. Because we close previous, the writer-hub might write something into that. I think it's ok right now.

#### What is the CloseIO and `StdinOnce`?

The contained-shim will open fifo twice for reading and writing. For the writing mode, the shim doesn't close stdin fifo until the client calls `CloseIO`. In some case, the pouch daemon might be crash before finishing the input. If shim doesn't hold writing mode fifo, the process in container will consider that it is EOF signal and exit. 

Based on this case, if the client sends EOF signal to input channel, the pouchd should send the CloseIO to shim to let the process exit.

`StdinOnce` in container's config is used by attach request. If the `StdinOnce` is true, when one attach request finishes stream copy, the pouchd will closes the input of process. So other attach requests to the same container will be stopped.

If the user wants `StdinOnce`, it should set it to true during creating container.
 
> NOTE: doc link is [here](https://docs.google.com/presentation/d/1TYC_W_uNKvmszKPhEqKXsbQXRKtiyAES3ulfaTsdhiM/edit?usp=sharing).

NONE

Added it.

Wait for CI.

I remove the `non-blocking` feature in this change and will add it in next patch.


fuweid:
ping @Starnop and @YaoZengzeng to help me review the cri part. Thanks

cli/run.go
```diff
@@ -71,6 +72,7 @@ func (rc *RunCommand) runRun(args []string) error {
 	}
 	containerName := rc.name
 	config.ContainerConfig.OpenStdin = rc.stdin
+	config.ContainerConfig.StdinOnce = rc.stdin // close stdin after attach connection closed
```
ZYecho: if pouch run with -i  -d flag enable and when the terminal detach, this container can't be attached again, right?

daemon/containerio/io.go
```diff
@@ -0,0 +1,178 @@
+package containerio
+
+import (
+	"io"
+	"time"
+
+	"github.com/alibaba/pouch/daemon/logger"
+	"github.com/alibaba/pouch/daemon/logger/crilog"
+	"github.com/alibaba/pouch/pkg/streams"
+	"github.com/sirupsen/logrus"
```
ZYecho: add blank line before this line?

test/cli_run_interactive_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// need to start command before write
+	cmd := command.PouchCmd("run",
+		"-i", "--net", "none",
+		"--name", name, busyboxImage, "cat")
+	cmd.Stdin = stdinR
+	cmd.Stdout = stdoutBuf
+	res := icmd.StartCmd(cmd)
+
+	// send bye to cat
```
ZYecho: bye -> byte

daemon/logger/copier.go
```diff
@@ -0,0 +1,81 @@
...
+		firstPartial = true
+		isPartial    bool
+		createdTime  time.Time
+
+		defaultBufSize = 16 * 1024
+	)
+
+	br := bufio.NewReaderSize(reader, defaultBufSize)
+	for {
+		bs, isPartial, err = br.ReadLine()
```
Starnop: ReadLine is a low-level line-reading primitive. Most callers should use ReadBytes('\n') or ReadString('\n') instead or use a Scanner. WDYT?

ZYecho:
In this pr desc,  the `stdinPipe` of the` Stream` struct was shared if there's more than one attacher? @fuweid  I did't know the difference between `stdinPipe` and `stdin` in `Stream` struct.

test/cli_run_interactive_test.go
fuweid: it's bye, not byte. LoL!

daemon/logger/copier.go
fuweid: what is difference? performance issue? or ?

cli/run.go
fuweid: it has been blocked by 

```
if (rc.attach || rc.stdin) && rc.detach { 
return fmt.Errorf("Conflicting options: -a (or -i) and -d")
}
``

Comment (fuweid) [reply cli/run.go]: > Maybe add a another flag for end-users to control is a good choice?

basically, I don't think so. The cli is just to help user to use pouch easier. This configuration is too complicated. I don't see any requirement for this.

File: daemon/logger/copier.go (hunk 4 above)
Comment (fuweid) [reply daemon/logger/copier.go]: synced with @Starnop . it can be solved.

File: cli/exec.go [hunk 5]
```diff
@@ -145,9 +146,12 @@ func holdHijackConnection(ctx context.Context, conn net.Conn, reader *bufio.Read
 	go func() {
 		if stdin {
 			io.Copy(conn, os.Stdin)
+			// close write if receive CTRL-D
+			if cw, ok := conn.(ioutils.CloseWriter); ok {
```
HusterWan: Do you mean `Ctr-D` is detach key?

fuweid: CTRL-D will send the EOF to close stdin. In here, we should close the writer not whole the connection, because we need to receive the data from reader.

Starnop:
I have tested the function `ReopenContainerLog`. Congratulations! Everything is ok, after this pull request merged, I will complete it.  And then, LGTM. :)

daemon/containerio/cio.go
```diff
@@ -2,229 +2,191 @@ package containerio
...
+	}
+
+	if withStdin {
+		cfg.Stdin = filepath.Join(fifoDir, processID+"-stdin")
 	}
 
-	if !stdin {
-		fifos.In = ""
+	if !withTerminal {
+		cfg.Stderr = filepath.Join(fifoDir, processID+"-stderr")
```
Ace-Tang: I do not think stderr fifo has related with terminal is open

daemon/containerio/cio.go
```diff
@@ -2,229 +2,191 @@ package containerio
...
+	for _, closer := range c.closers {
+		if closer == nil {
+			continue
+		}
+		if err := closer.Close(); err != nil {
+			lastErr = err
+		}
 	}
-	if g.dir != "" {
-		return os.RemoveAll(g.dir)
```
Ace-Tang: Before, we remove fifo dir here, now it moved to closeFn, does it matter

daemon/containerio/cio.go
```diff
@@ -2,229 +2,191 @@ package containerio
...
+	if c.cancel != nil {
+		c.cancel()
 	}
-	return nil
 }
 
-func (g *wgCloser) Cancel() {
-	g.cancel()
+// TODO(fuweid): pipes will be removed when update vendor to containerd@v1.2.
+type pipes struct {
```
Ace-Tang: Do we use pipe, direct IO before ? I will do more check

daemon/containerio/cio.go
```diff
@@ -2,229 +2,191 @@ package containerio
...
+		containerIO: containerIO{
+			config:  fifos.Config,
+			closers: append(pipes.Closers(), fifos),
+			cancel:  cancel,
+		},
+	}, nil
 }
 
-func copyIO(fifos *containerdio.FIFOSet, ioset *ioSet, tty bool) (_ *wgCloser, err error) {
+func openPipes(ctx context.Context, fifos *CioFIFOSet) (_ pipes, err0 error) {
```
Ace-Tang: I know, you implement pipe for here use.

daemon/containerio/cio.go
fuweid: Yes. I checked that containerd-shim. The shim will open socket as pty-slave. There is only one output from pty-slave so that we don't need stderr fifo if the `withTerminal` is true

rudyfly:
Remove blank line https://github.com/alibaba/pouch/pull/2375/files#diff-d9e51acfdfc53f5e42e18981e1d9f2b1R712

fuweid:
any update @Ace-Tang @YaoZengzeng

ctrd/container.go
```diff
@@ -791,3 +831,107 @@ func withCheckpointOpt(checkpoint *containerdtypes.Descriptor) containerd.NewTas
...
+			stdinOnce sync.Once
+		)
+		oldStdin := cdio.Stdin
+		cdio.Stdin = ioutils.NewWriteCloserWrapper(oldStdin, func() error {
+			stdinOnce.Do(func() {
+				errClose = oldStdin.Close()
+
+				// Both the caller and container/exec process holds write side pipe
+				// for the stdin. When the caller closes the write pipe, the process doesn't
+				// exit until the caller.
```
YaoZengzeng: Maybe we missed some comment here? 😆

fuweid:
@YaoZengzeng I have updated that the `containerio` does `Close` action which includes `Wait` action. However, just in case, we wait for the Wait in 10 seconds. If not, we will go ahead.

pkg/streams/multi.go
```diff
@@ -0,0 +1,48 @@
...
+			w.Close()
+			evictIdx = append(evictIdx, n)
+		}
+	}
+
+	for n, i := range evictIdx {
+		mw.writers = append(mw.writers[:i-n], mw.writers[i-n+1:]...)
+	}
+	mw.Unlock()
+	return len(p), nil
```
zhuangqh: if all of the writer writing bytes fails here, still return `len(p)`. Is it reasonable?

fuweid: since we don't return error here, we need to return the len(p) here. The writer hub uses io.Pipe as `io.WriteCloser`. All data are handled in memory, not the real file or network IO. If it run into a error, there is some wrong with pouch and we should restart pouchd.

I think I can add the warning message here if it run into error during writing. How do you think?

pkg/streams/utils.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// CopyPipes will watchs the data pipe's channel, like sticked to the pipe.
+//
+// NOTE: don't assign the specific type to the Pipes because the Std* != nil
+// always return true.
+func (s *Stream) CopyPipes(p Pipes) {
+	copyfn := func(styp string, w io.WriteCloser, r io.ReadCloser) {
+		s.Add(1)
+		go func() {
```
zhuangqh: Controlled by waitGroup to promise all of the data has finished copying(abnormal exit is also regarded as finish). Just comment to record this.

pkg/streams/utils.go
```diff
@@ -0,0 +1,185 @@
...
+	if p.Stdout != nil {
+		copyfn("stdout", s.Stdout(), p.Stdout)
+	}
+
+	if p.Stderr != nil {
+		copyfn("stderr", s.Stderr(), p.Stderr)
+	}
+
+	if s.stdin != nil && p.Stdin != nil {
+		go func() {
```
zhuangqh: Not controlled by waitGroup. Because it may be suddenly closed by user. Just comment to record this.

apis/server/container_bridge.go
```diff
@@ -284,26 +286,40 @@ func (s *Server) renameContainer(ctx context.Context, rw http.ResponseWriter, re
...
+	// close hijack stream
+	defer closeFn()
+
+	if upgrade {
+		fmt.Fprintf(stdout, "HTTP/1.1 101 UPGRADED\r\nContent-Type: application/vnd.docker.raw-stream\r\nConnection: Upgrade\r\nUpgrade: tcp\r\n\r\n")
+	} else {
+		fmt.Fprintf(stdout, "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.docker.raw-stream\r\n\r\n")
 	}
 
+	attach.UseStdin = req.FormValue("stdin") == "1"
```
zhuangqh: using `httputils.BoolValue(req, "stdin")` instead?

fuweid:
ping @zhuangqh I update with your recommendation. please take a look.

YaoZengzeng:
LGTM
@fuweid Your work is really brilliant 😎

ctrd/container.go
```diff
@@ -791,3 +831,107 @@ func withCheckpointOpt(checkpoint *containerdtypes.Descriptor) containerd.NewTas
 		return nil
 	}
 }
+
+// InitStdio allows caller to handle any initialize job.
```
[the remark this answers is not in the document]
fuweid: I think job is ok :joy:

daemon/mgr/container.go
```diff
@@ -1537,77 +1559,54 @@ func (mgr *ContainerManager) updateNetworkSettings(container *Container, n libne
...
+		cntrio *containerio.IO
+		err    error
+	)
 
-	if attach != nil {
-		options = append(options, attachConfigToOptions(attach)...)
-		options = append(options, containerio.WithStdin(attach.Stdin))
-	} else {
-		options = append(options, containerio.WithDiscard())
+	if cntrio = mgr.IOs.Get(c.ID); cntrio == nil {
```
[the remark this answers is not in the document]
fuweid: container -> cntr I think it's ok 😅

daemon/containerio/cio.go
```diff
@@ -2,229 +2,191 @@ package containerio
...
-	if !stdin {
-		fifos.In = ""
+	if !withTerminal {
+		cfg.Stderr = filepath.Join(fifoDir, processID+"-stderr")
 	}
 
-	return fifos, nil
+	closeFn := func() error {
+		if err := os.RemoveAll(fifoDir); err != nil {
+			logrus.WithError(err).Warnf("failed to remove process(id=%v) fifo dir", processID)
```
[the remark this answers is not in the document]
fuweid: got it.

cri/v1alpha2/cri.go
```diff
@@ -27,6 +26,7 @@ import (
 	"github.com/alibaba/pouch/pkg/errtypes"
 	"github.com/alibaba/pouch/pkg/meta"
 	"github.com/alibaba/pouch/pkg/reference"
+	pkgstreams "github.com/alibaba/pouch/pkg/streams"
```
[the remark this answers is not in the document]
rudyfly: cri package is `steam`, your package is `steams`, he he he he

fuweid:
ping @zhuangqh @rudyfly I changes cntrio -> ctrio in containerio package.

PR #2392, opened 2018-10-31
feature: add Close function for boltdb backend

HusterWan:

When testing [d2p-migrator](https://github.com/pouchcontainer/d2p-migrator), I found I need close the `boltdb` store after used, so that the `pouchd` can take charge of the  volume store.

So I add a `Shutdown` function for the `Store`

none

add `TestBoltdbClose` case

none

none

### DIFF
--- a/pkg/meta/store.go
+++ b/pkg/meta/store.go
@@ -270,3 +270,8 @@ func (s *Store) KeysWithPrefix(prefix string) ([]string, error) {
 func (s *Store) Path(key string) string {
 	return s.backend.Path(key)
 }
+
+// Shutdown releases all resources used by the backend
+func (s *Store) Shutdown() error {
+	return s.backend.Close()
+}
--- a/pkg/meta/store_test.go
+++ b/pkg/meta/store_test.go
@@ -1,8 +1,13 @@
 package meta
 
 import (
+	"fmt"
+	"os"
+	"path"
 	"reflect"
 	"testing"
+
+	"github.com/alibaba/pouch/pkg/utils"
 )
 
 type Demo struct {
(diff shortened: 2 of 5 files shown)


pkg/meta/store.go
```diff
@@ -270,3 +270,8 @@ func (s *Store) KeysWithPrefix(prefix string) ([]string, error) {
 func (s *Store) Path(key string) string {
 	return s.backend.Path(key)
 }
+
+// Shutdown releases all resources used by the backend
+func (s *Store) Shutdown() error {
```
rudyfly: Where to call the function `Shutdown`？

HusterWan: I need it in `d2p-migrator` :)

pkg/meta/store_test.go
```diff
@@ -201,6 +202,18 @@ func TestBoltdbRemove(t *testing.T) {
 	}
 }
 
+func TestBoltdbClose(t *testing.T) {
+	if err := boltdbStore.Shutdown(); err != nil {
```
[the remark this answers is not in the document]
HusterWan: Just stay with the origin test cases, but i think your advice is reasonable, i will update it

pkg/meta/store_test.go
```diff
@@ -117,20 +208,41 @@ func (d *Demo3) Key() string {
...
-	Buckets: []Bucket{
-		{"boltdb", reflect.TypeOf(Demo3{})},
-	},
+	return NewStore(boltdbCfg)
+}
+
+func ensureFileNotExist(file string) error {
+	_, err := os.Stat(file)
+	if err == nil {
+		os.Remove(file)
```
zhuangqh: Actually, `file` is a directory path, using `os.RemoveAll` instead.

pkg/meta/store_test.go
```diff
@@ -93,16 +145,55 @@ func TestFetch(t *testing.T) {
...
 	}
 }
 
 func TestRemove(t *testing.T) {
+	// initialize
+	dbFile := path.Join("/tmp", utils.RandString(8, "TestRemove", ""))
+	if err := ensureFileNotExist(dbFile); err != nil {
+		t.Fatal(err)
+	}
+	defer ensureFileNotExist(dbFile)
```
zhuangqh: What about creating a test wrapper? like
```go
func storeTestWrapper(t *testing.T, name string, test func(*Store)) {
	dbDir := path.Join("/tmp", utils.RandString(8, name, ""))
	if err := ensureFileNotExist(dbDir); err != nil {
		t.Fatal(err)
	}
	defer ensureFileNotExist(dbDir)

	s, err := initLocalStore(dbDir)
	if err != nil {
		t.Fatal(err)
	}

	test(s)
}
```
```go
func TestGet(t *testing.T) {
	storeTestWrapper(t, "TestGet", func(s *Store) {
              // xxx
	})
}
```

Don't repeat yourself

pkg/meta/store_test.go
HusterWan: I am afraid, `file` is a file path

pkg/meta/store_test.go
HusterWan: great advice

zhuangqh:
LGTM

rudyfly:
LGTM

PR #2115, opened 2018-08-19
 refactor: refactor security part in create spec 

Ace-Tang:
### Ⅰ. Describe what this PR did

1. security parameter include seccomp, apparmor and no-new-privileges
are not implement good, refactor these part when create new spec.

2. fix device resource add two times.

3. use pouch default runtime spec, compare to containerd spec, we remove
Root, Process(Env Cmd NoNewPrivileges User Rlimits), Mount(/run),
Linux(CgroupsPath), we add more allowed device, add cgroup mount.

PR #2411, opened 2018-11-02
oci: add pouch default runtime spec

### LINKED ISSUE #2116: [refactor] create runtime spec need refactor
### Ⅰ. Issue Description

Many part in runtime spec is not implemented good in pouch, here is a list we need to re-consider.
* security part, include seccomp, apparmor and no-new-priviledges
* shm, for shared filesystem

1.
2.
3.

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Ace-Tang:
use pouch default runtime spec, instead of containerd default spec,
compare to containerd spec, we remove Root, Process(Env Cmd
NoNewPrivileges User Rlimits), Linux(CgroupsPath), this should not
be exist in default spec, remove Mount(/run), this is not used, add
allowed device, add cgroup mount.

use pouch default runtime spec, instead of containerd default spec,
compare to containerd spec, we remove Root, Process(Env Cmd
NoNewPrivileges User Rlimits), Linux(CgroupsPath), this should not
be exist in default spec, remove Mount(/run), this is not used, add
allowed device, add cgroup mount.

fix #2116 , and base on PR #2271 ， close #2115 

no.

For the default runtime spec pouch used, I will post an explanation, if have doubt with the new spec , please wait some time.

### DIFF
--- a/oci/spec_default.go
+++ b/oci/spec_default.go
@@ -1,0 +1,187 @@
+package oci
+
+import (
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func iPtr(i int64) *int64 { return &i }
+
+func defaultCaps() []string {
+	return []string{
+		"CAP_CHOWN",
+		"CAP_DAC_OVERRIDE",
+		"CAP_FOWNER",
+		"CAP_FSETID",
+		"CAP_KILL",
+		"CAP_SETGID",
+		"CAP_SETUID",
+		"CAP_SETPCAP",
+		"CAP_AUDIT_WRITE",
+		"CAP_NET_BIND_SERVICE",
+		"CAP_NET_RAW",
+		"CAP_SYS_CHROOT",
+		"CAP_MKNOD",
+		"CAP_SETFCAP",
+	}
+}
+
+// NewDefaultSpec create a new spec.
+func NewDefaultSpec() *specs.Spec {
+	s := &specs.Spec{
+		Version: specs.Version,
+	}
+	s.Mounts = []specs.Mount{
+		{
+			Destination: "/proc",
+			Type:        "proc",
+			Source:      "proc",
+			Options:     []string{"nosuid", "noexec", "nodev"},
+		},
+		{
+			Destination: "/dev",
+			Type:        "tmpfs",
+			Source:      "tmpfs",
+			Options:     []string{"nosuid", "strictatime", "mode=755"},
+		},
+		{
+			Destination: "/dev/pts",
+			Type:        "devpts",
+			Source:      "devpts",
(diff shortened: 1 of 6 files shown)


oci/spec_default.go
```diff
@@ -0,0 +1,194 @@
+package oci
```
rudyfly: too many packages in pouch directory，can we merge them？

oci/spec_default.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+		{
+			Destination: "/dev/shm",
+			Type:        "tmpfs",
+			Source:      "shm",
+			Options:     []string{"nosuid", "noexec", "nodev", "mode=1777", "size=65536k"},
+		},
+	}
+
+	if _, err := os.Stat("/proc/self/ns/net"); err == nil {
```
fuweid: could you help to explain reason that why we should add the check here?

fuweid: @Ace-Tang I think we should remove this part and add cgroup directly. cc @rudyfly

Ace-Tang: Agree that

fuweid: @Ace-Tang could you help to update the PR? thanks

oci/spec_default.go
```diff
@@ -0,0 +1,187 @@
...
+			Options:     []string{"nosuid", "noexec", "nodev", "ro"},
+		},
+		{
+			Destination: "/dev/mqueue",
+			Type:        "mqueue",
+			Source:      "mqueue",
+			Options:     []string{"nosuid", "noexec", "nodev"},
+		},
+		{
+			Destination: "/dev/shm",
```
rudyfly: Other method: in host mounting `tmpfs` to one directory and bind it into container.
Any different from this way？

rudyfly:
LGTM
I have checked `capabilities` and `mounts`, they are right.

oci/spec_default.go
Ace-Tang: I want to remove /dev/shm from default spec, since the option is related with what kind of ipc ns container choose, it trouble to change option.

fuweid:
@rudyfly, any update here?

rudyfly:
@fuweid yes， it need update. I move LGTM

rudyfly:
LGTM

Ace-Tang:
Update with fix
```
if sm.Destination == "/dev/shm" && c.HostConfig.ShmSize != nil &&
+                       *c.HostConfig.ShmSize != 0 {
+                       for idx, v := range sm.Options {
+                               if strings.Contains(v, "size=") {
+                                       sm.Options[idx] = fmt.Sprintf("size=%s", strconv.FormatInt(*c.HostConfig.ShmSize, 10))
+                               }
+                       }
                }
```
and add a test for default shm size, @rudyfly

PR #2428, opened 2018-11-05
feature: use ringBuffer to make log not blocking and make it configurable

Starnop:

As the title describes.

None.

Add it Later.

Add it Later.

### DIFF
--- a/daemon/logger/logbuffer/logbuff.go
+++ b/daemon/logger/logbuffer/logbuff.go
@@ -1,0 +1,61 @@
+package logbuffer
+
+import (
+	"github.com/alibaba/pouch/daemon/logger"
+
+	"github.com/sirupsen/logrus"
+)
+
+// LogBuffer is uses to cache the container's logs with ringBuffer.
+type LogBuffer struct {
+	ringBuffer *RingBuffer
+	logger     logger.LogDriver
+}
+
+// NewLogBuffer return a new BufferLog.
+func NewLogBuffer(logDriver logger.LogDriver, maxBytes int64) (logger.LogDriver, error) {
+	bl := &LogBuffer{
+		logger:     logDriver,
+		ringBuffer: NewRingBuffer(maxBytes),
+	}
+
+	// use a goroutine to write logs continuously with specified log driver
+	go bl.run()
+	return bl, nil
+}
+
+// Name return the log driver's name.
+func (bl *LogBuffer) Name() string {
+	return bl.logger.Name()
+}
+
+// WriteLogMessage will write the LogMessage to the ringBuffer.
+func (bl *LogBuffer) WriteLogMessage(msg *logger.LogMessage) error {
+	return bl.ringBuffer.Push(msg)
+}
+
+// Close close the ringBuffer and drain the messages.
+func (bl *LogBuffer) Close() error {
+	bl.ringBuffer.Close()
+	for _, msg := range bl.ringBuffer.Drain() {
+		if err := bl.logger.WriteLogMessage(msg); err != nil {
+			logrus.Debugf("failed to write log %v when closing with log driver %s", msg, bl.logger.Name())
+		}
+	}
+
+	return bl.logger.Close()
+}
+
+// write logs continuously with specified log driver from ringBuffer.
+func (bl *LogBuffer) run() {
+	for {
+		msg, err := bl.ringBuffer.Pop()
+		if err != nil {
+			return
+		}
+
+		if err := bl.logger.WriteLogMessage(msg); err != nil {
+			logrus.Debugf("failed to write log %v with log driver %s", msg, bl.logger.Name())
+		}
+	}
+}
--- a/daemon/logger/logbuffer/ringbuff.go
+++ b/daemon/logger/logbuffer/ringbuff.go
@@ -1,35 +1,41 @@
-package ringbuffer
+package logbuffer
 
 import (
 	"fmt"
 	"sync"
+
+	"github.com/alibaba/pouch/daemon/logger"
 )
 
 // ErrClosed is used to indicate the ringbuffer has been closed.
 var ErrClosed = fmt.Errorf("closed")
 
-const defaultSize = 1024
+const (
+	defaultMaxBytes = 1e6 //1MB
+)
 
 // RingBuffer implements a fixed-size buffer which will drop oldest data if full.
 type RingBuffer struct {
 	mu   sync.Mutex
 	wait *sync.Cond
 
-	cap    int
 	closed bool
 	q      *queue
+
+	maxBytes     int64
+	currentBytes int64
 }
 
-// New creates new RingBuffer.
-func New(cap int) *RingBuffer {
-	if cap <= 0 {
-		cap = defaultSize
+// NewRingBuffer creates new RingBuffer.
+func NewRingBuffer(maxBytes int64) *RingBuffer {
+	if maxBytes < 0 {
+		maxBytes = defaultMaxBytes
 	}
 
 	rb := &RingBuffer{
-		cap:    cap,
-		closed: false,
-		q:      newQueue(),
+		closed:   false,
+		q:        newQueue(),
+		maxBytes: maxBytes,
 	}
 	rb.wait = sync.NewCond(&rb.mu)
 	return rb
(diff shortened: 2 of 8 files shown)


daemon/mgr/container_logger.go
```diff
@@ -5,27 +5,46 @@ import (
...
 		logrus.Warnf("not support (%v) log driver yet", cfg.LogDriver)
 		return nil, nil
 	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	if info.LogConfig["mode"] == LogNonBlocking {
+		return bufferlog.NewBufferLog(ld, info)
```
fuweid: could we postpone to start the goroutine after startLogging?

daemon/logger/bufferlog/bufferlog.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return err
+}
+
+// Close close the ringBuffer and drain the messages.
+func (bl *BufferLog) Close() error {
+	bl.ringBuffer.Close()
+	var hasErr bool
+	for _, msg := range bl.ringBuffer.Drain() {
+		msgTemp, ok := msg.(*logger.LogMessage)
```
fuweid: I think we should not use `interface{}` here because the ringbuffer is only used by logging. So if we do, there is no need to assert the value each time.

daemon/logger/bufferlog/bufferlog.go
```diff
@@ -0,0 +1,93 @@
...
+
+// BufferLog is uses to cache the container's logs with ringBuffer.
+type BufferLog struct {
+	ringBuffer *rb.RingBuffer
+	logger     logger.LogDriver
+}
+
+// NewBufferLog return a new BufferLog.
+func NewBufferLog(logDriver logger.LogDriver, info logger.Info) (logger.LogDriver, error) {
+	bufferSize := int64(-1)
```
HusterWan: Just curious, why should you define the `bufferSize` to `int`? 

So, you no need covert the `int64` to `int` in line 31

daemon/logger/bufferlog/bufferlog.go
```diff
@@ -0,0 +1,93 @@
...
+// Name return the log driver's name.
+func (bl *BufferLog) Name() string {
+	return bl.logger.Name()
+}
+
+// WriteLogMessage will write the LogMessage to the ringBuffer.
+func (bl *BufferLog) WriteLogMessage(msg *logger.LogMessage) error {
+	covered, err := bl.ringBuffer.Push(msg)
+	if covered {
+		logrus.Warnf("data coverage occurs: %v", msg)
```
HusterWan: Please change the log level from `Warnf` to `Debugf`, because containers may produce too many this log

daemon/logger/bufferlog/bufferlog.go
```diff
@@ -0,0 +1,93 @@
...
+	bl.ringBuffer.Close()
+	var hasErr bool
+	for _, msg := range bl.ringBuffer.Drain() {
+		msgTemp, ok := msg.(*logger.LogMessage)
+		if !ok {
+			// TODO anything else should be done?
+			logrus.Warnf("failed to do type assertion when closing: %v", msg)
+			continue
+		}
+		if hasErr {
```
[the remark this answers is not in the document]
fuweid: I don't think we should put it back because that data is allowed to be missed with non-blocking mode.

daemon/logger/bufferlog/bufferlog.go
```diff
@@ -0,0 +1,93 @@
+package bufferlog
```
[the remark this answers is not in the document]
ZYecho: Maybe the package name logBuffer is more accurate？ since the emphasis is buffer.

daemon/mgr/container_logger.go
```diff
@@ -5,27 +5,46 @@ import (
...
 	default:
 		logrus.Warnf("not support (%v) log driver yet", cfg.LogDriver)
 		return nil, nil
 	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	if info.LogConfig["mode"] == LogNonBlocking {
```
ZYecho: what here need validate logic for this log Opt? since the mode can be `blocking` or anything else.

daemon/logger/bufferlog/bufferlog.go
Starnop: Agree with that.

daemon/logger/bufferlog/bufferlog.go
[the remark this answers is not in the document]
Starnop: Maybe add a number such as three for retry, WDYT? @HusterWan @fuweid

daemon/logger/bufferlog/bufferlog.go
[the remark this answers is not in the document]
Starnop: @ZYecho In favor of it. Thanks.

daemon/mgr/container_logger.go
Starnop: Yes, in default, there is no buffer that will result in blocking if  there is a lot of data.

Starnop:
@HusterWan @fuweid All mentioned above has been updated. PTAL.

daemon/logger/logbuffer/logbuff.go
```diff
@@ -0,0 +1,65 @@
...
+type LogBuffer struct {
+	ringBuffer *RingBuffer
+	logger     logger.LogDriver
+}
+
+// NewLogBuffer return a new BufferLog.
+func NewLogBuffer(logDriver logger.LogDriver, maxBytes int64) (logger.LogDriver, error) {
+	bl := &LogBuffer{
+		logger:     logDriver,
+		ringBuffer: New(-1, maxBytes),
```
fuweid: could we use other name here, since the `New` contains less information?

daemon/logger/logbuffer/ringbuff.go
```diff
@@ -18,26 +23,33 @@ type RingBuffer struct {
...
 	closed bool
 	q      *queue
+
+	maxBytes     int64
+	currentBytes int64
 }
 
 // New creates new RingBuffer.
-func New(cap int) *RingBuffer {
+func New(cap int, maxBytes int64) *RingBuffer {
```
fuweid: I think we should remove the `cap` here since the buffer is controlled by the length of bytes, not `cap` here.

daemon/logger/logbuffer/ringbuff.go
```diff
@@ -18,26 +23,33 @@ type RingBuffer struct {
...
+		maxBytes: maxBytes,
 	}
 	rb.wait = sync.NewCond(&rb.mu)
 	return rb
 }
 
 // Push pushes value into buffer and return whether it covers the oldest data
 // or not.
-func (rb *RingBuffer) Push(val interface{}) (bool, error) {
+func (rb *RingBuffer) Push(val *logger.LogMessage) (bool, error) {
```
fuweid: useless cover return-value here

Starnop: I agree with you, and the mentioned above has been updated. :)

fuweid:
Basically, LGTM. But, @Starnop could you help to add benchmark cases for the logbuffer?

daemon/mgr/container.go
```diff
@@ -1582,10 +1588,22 @@ func (mgr *ContainerManager) initLogDriverBeforeStart(c *Container) error {
...
+	logDriver, err := logOptionsForContainerio(c, logInfo)
 	if err != nil {
 		return err
 	}
+
+	if logInfo.LogConfig["mode"] == LogNonBlocking {
+		if maxBufferSize, ok := logInfo.LogConfig["max-buffer-size"]; ok {
+			maxBytes, err := units.RAMInBytes(maxBufferSize)
+			if err != nil {
+				return errors.Wrap(err, "failed to  parse option max-buffer-size")
```
fuweid: remove the additional space

Starnop:
> Basically, LGTM. But, @Starnop could you help to add benchmark cases for the logbuffer?

Maybe  add benchmark cases in the next PR?

daemon/mgr/container.go
Starnop: good catch, Thank you!

fuweid:
oops. @Starnop could you help to take a look for this?

```
$ sudo pouch run --rm --log-opt mode=non-blocking busybox sh -c 'for i in $(seq 1 1000); do echo 1; done' | wc -l
Error: failed to run container: {"message":"unknown log opt 'mode' for json-file log driver"}
```

fuweid:
> Maybe add benchmark cases in the next PR?

Sure.

daemon/mgr/container_validation.go
```diff
@@ -217,9 +219,41 @@ func (mgr *ContainerManager) validateLogConfig(c *Container) error {
...
+
+		// try to parse the max-buffer-size option
+		_, err := units.RAMInBytes(maxBufferSize)
+		if err != nil {
+			return errors.Wrapf(err, "failed to parse option max-buffer-size: %s", maxBufferSize)
+		}
+	}
+
+	// filter the option which have been validated such as mode, max-buffer-size.
+	var filteredOpts = map[string]bool{
```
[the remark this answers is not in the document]
Starnop: Yeah, has been updated.

daemon/mgr/container_validation.go
```diff
@@ -27,6 +29,12 @@ var (
 
 	errInvalidDevice = errors.New("invalid nvidia device")
 	errInvalidDriver = errors.New("invalid nvidia driver capability")
+
+	// filteredOpts the option which should be validated in common such as mode, max-buffer-size.
+	filteredOpts = map[string]bool{
```
fuweid: could we change other name here? this name is confusing.

PR #2443, opened 2018-11-07
bugfix: use netNSPath dynamically

Starnop:

The sandbox will be stopped due to various reasons ,such as the physical machine restart, if perform delete operations at this time which will cause the failure of network IP release. The reason for this is that can't find netnspath  because the pid has changed and the cache has not been updated. Eventually leading to pod has been terminating and IP leakage.
So instead of using cache, we use container's pid to splice netnspath which will ensure that the netNSPath is correct.

None.

None.

### DIFF
--- a/cri/ocicni/cni_manager.go
+++ b/cri/ocicni/cni_manager.go
@@ -8,6 +8,7 @@ import (
 
 	cnicurrent "github.com/containernetworking/cni/pkg/types/current"
 	"github.com/cri-o/ocicni/pkg/ocicni"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
@@ -75,11 +76,18 @@ func (c *CniManager) SetUpPodNetwork(podNetwork *ocicni.PodNetwork) error {
 
 // TearDownPodNetwork is the method called before a pod's sandbox container will be deleted.
 func (c *CniManager) TearDownPodNetwork(podNetwork *ocicni.PodNetwork) error {
+	// perform the teardown network operation whatever to
+	// give CNI Plugin a chance to perform some operations
 	err := c.plugin.TearDownPod(*podNetwork)
-	if err != nil {
-		return fmt.Errorf("failed to destroy network for sandbox %q: %v", podNetwork.ID, err)
+	if err == nil {
+		return nil
 	}
-	return nil
+
+	// if netNSPath is not found, should return the error of IsNotExist.
+	if _, err = os.Stat(podNetwork.NetNS); err != nil {
+		return err
+	}
+	return errors.Wrapf(err, "failed to destroy network for sandbox %q", podNetwork.ID)
 }
 
 // GetPodNetworkStatus is the method called to obtain the ipv4 or ipv6 addresses of the pod sandbox.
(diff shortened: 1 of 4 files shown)


cri/ocicni/cni_manager.go
```diff
@@ -77,6 +77,9 @@ func (c *CniManager) SetUpPodNetwork(podNetwork *ocicni.PodNetwork) error {
 func (c *CniManager) TearDownPodNetwork(podNetwork *ocicni.PodNetwork) error {
 	err := c.plugin.TearDownPod(*podNetwork)
 	if err != nil {
+		if _, err = os.Stat(podNetwork.NetNS); err != nil {
```
fengzixu: There is a more clear codestyle:

```
if err == nil {
      return nil
}

if _, err = os.Stat(podNetwork.NetNS); err != nil {
          return err
}
return fmt.Errorf("failed to destroy network for sandbox %q: %v", podNetwork.ID, err)
```
Decrease the nested layers is a quie important rule in clean codes

cri/v1alpha2/cri_utils.go
```diff
@@ -473,10 +473,10 @@ func (c *CriManager) setupPodNetwork(ctx context.Context, id string, config *run
 		PortMappings: toCNIPortMappings(config.GetPortMappings()),
 	})
 	if err != nil {
-		return "", err
+		return err
```
fengzixu: It seems like that there is no need to check error single. A recommended format:

```go
if err := err = c.CniMgr.SetUpPodNetwork(&ocicni.PodNetwork{
        PortMappings: toCNIPortMappings(config.GetPortMappings()),
}); err != nil {
      return err
}
```

cri/ocicni/cni_manager.go
```diff
@@ -77,6 +77,9 @@ func (c *CniManager) SetUpPodNetwork(podNetwork *ocicni.PodNetwork) error {
...
+			return err
+		}
 		return fmt.Errorf("failed to destroy network for sandbox %q: %v", podNetwork.ID, err)
```
rudyfly: use `errors.Wrapf` to wrap error

cri/ocicni/cni_manager.go
rudyfly: agree

Starnop: Agree with that, THX.

cri/ocicni/cni_manager.go
```diff
@@ -76,10 +77,14 @@ func (c *CniManager) SetUpPodNetwork(podNetwork *ocicni.PodNetwork) error {
 // TearDownPodNetwork is the method called before a pod's sandbox container will be deleted.
 func (c *CniManager) TearDownPodNetwork(podNetwork *ocicni.PodNetwork) error {
 	err := c.plugin.TearDownPod(*podNetwork)
```
HusterWan: how about:
```
if err :=  c.plugin.TearDownPod(*podNetwork); err == nil {
    return nil
}
```

Starnop: `err` will be used later. So no need to change. :)

rudyfly:
LGTM

cri/ocicni/cni_manager.go
HusterWan: make sense

PR #2447, opened 2018-11-07
feature: implement the feature of reopen log about cri

Starnop:

Thanks for [the PR of IO redesign ](https://github.com/alibaba/pouch/pull/2375) , we could implement the function of `ReopenContainerLog`.

BTW, delete the skip test about `ReopenContainerLog` .

None.

None.

use critest
```
critest --runtime-endpoint /var/run/pouchcri.sock --ginkgo.focus="runtime should support reopening container log"
```


cri/v1alpha2/cri.go
```diff
@@ -1161,7 +1189,30 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
 func (c *CriManager) ReopenContainerLog(ctx context.Context, r *runtime.ReopenContainerLogRequest) (*runtime.ReopenContainerLogResponse, error) {
-	return nil, fmt.Errorf("ReopenContainerLog Not Implemented Yet")
+	containerID := r.GetContainerId()
+
+	container, err := c.ContainerMgr.Get(ctx, containerID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get container %q with error: %v", containerID, err)
+	}
+	if !container.IsRunning() {
+		return nil, fmt.Errorf("container is not running")
```
fuweid: I was thinking that CRI should use the right status code for the error. For this one, we should use ErrPreconditionFail something like that. How do you think about this?

cri/v1alpha2/cri.go
```diff
@@ -180,6 +183,20 @@ func NewCriManager(config *config.Config, ctrMgr mgr.ContainerMgr, imgMgr mgr.Im
 		return nil, fmt.Errorf("failed to create sandbox meta store: %v", err)
 	}
 
+	c.ContainerStore, err = meta.NewStore(meta.Config{
```
fuweid: could you provide more detail about this design? I think it should be described in the PR.

cri/v1alpha2/cri.go
```diff
@@ -532,28 +549,29 @@ func (c *CriManager) RemovePodSandbox(ctx context.Context, r *runtime.RemovePodS
 
 	// Remove all containers in the sandbox.
 	for _, container := range containers {
-		err = c.ContainerMgr.Remove(ctx, container.ID, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true})
-		if err != nil {
+		if err = c.ContainerMgr.Remove(ctx, container.ID, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true}); err != nil {
 			return nil, fmt.Errorf("failed to remove container %q of sandbox %q: %v", container.ID, podSandboxID, err)
 		}
+
+		if err = c.ContainerStore.Remove(container.ID); err != nil {
```
HusterWan: werid

cri/v1alpha2/cri.go
Starnop: Yeah, Agree with that. THX.

cri/v1alpha2/cri.go
Starnop: Maybe a mapping is better. I will redesign it. THX.

cri/v1alpha2/cri.go
```diff
@@ -257,11 +258,15 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
...
+
 	// once sandbox container created, we are obligated to store it.
 	sandboxMeta := &SandboxMeta{
-		ID:      id,
-		Config:  config,
-		Runtime: config.Annotations[anno.KubernetesRuntime],
+		ID:              id,
+		Config:          config,
+		Runtime:         config.Annotations[anno.KubernetesRuntime],
+		ContainerLogMap: containerLogMap,
```
fuweid: could we use `ContainerLogMap: make(map[string]string)` to do initialization? since the `containerLogMap` is only used for assignment.

cri/v1alpha2/cri.go
```diff
@@ -758,6 +761,12 @@ func (c *CriManager) CreateContainer(ctx context.Context, r *runtime.CreateConta
 		if err := c.ContainerMgr.AttachCRILog(ctx, containerID, logPath); err != nil {
 			return nil, err
 		}
+
+		sandboxMeta.ContainerLogMap[containerID] = logPath
```
fuweid: could we update the data before `AttachCRILog`?

Starnop:
All mentioned above have been updated. PTAL. THX.

cri/v1alpha2/cri.go
```diff
@@ -765,6 +764,11 @@ func (c *CriManager) CreateContainer(ctx context.Context, r *runtime.CreateConta
 	// Get container log.
 	if config.GetLogPath() != "" {
 		logPath := filepath.Join(sandboxConfig.GetLogDirectory(), config.GetLogPath())
+		sandboxMeta.ContainerLogMap[containerID] = logPath
```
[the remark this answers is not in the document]
Starnop: THX, I have updated it.

cri/v1alpha2/cri.go
```diff
@@ -1128,7 +1132,41 @@ func (c *CriManager) UpdateContainerResources(ctx context.Context, r *runtime.Up
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get metadata of %q from SandboxStore: %v", podSandboxID, err)
+	}
+	sandboxMeta, ok := res.(*SandboxMeta)
+	if !ok {
+		return nil, fmt.Errorf("failed to type asseration for sandboxMeta: %v", res)
+	}
+	logPath, ok := sandboxMeta.ContainerLogMap[containerID]
+	if !ok {
+		logrus.Infof("log path of container: %q is empty", containerID)
```
zhuangqh: Why did you just print a log here. Empty logPath is not acceptable in `attachCRILog`.
Please also add a precheck at https://github.com/alibaba/pouch/blob/master/daemon/mgr/container.go#L975 
if logPath is empty, immediately return error

cri/v1alpha2/cri.go
```diff
@@ -499,28 +501,26 @@ func (c *CriManager) RemovePodSandbox(ctx context.Context, r *runtime.RemovePodS
 
 	// Remove all containers in the sandbox.
 	for _, container := range containers {
-		err = c.ContainerMgr.Remove(ctx, container.ID, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true})
-		if err != nil {
+		if err := c.ContainerMgr.Remove(ctx, container.ID, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true}); err != nil {
 			return nil, fmt.Errorf("failed to remove container %q of sandbox %q: %v", container.ID, podSandboxID, err)
```
fuweid: I think we can use `errors.Wrap` here because the `Remove` might return the 404 or specific error.

We can handle this in next pr. it is not blocked issue.

fuweid:
Two things should be done after this PR:

1. use `errors.Wrap` in the gRPC service level because we can use `Cause` to check the specific error type.
2. Missing recovery after restart pouch daemon

cc @Starnop @zhuangqh

PR #2451, opened 2018-11-08
bugfix: use a fixed version of cni instead of the branch of master

Starnop:

We use the master branch of `containernetworking/plugins` when install cni , with this PR https://github.com/containernetworking/plugins/pull/219 submitted, we got error info when running the test about cri
```
install cni...
hack/install/install_cni.sh: line 26: /home/travis/gopath/src/github.com/containernetworking/plugins/build.sh: No such file or directory
make: *** [cri-v1alpha1-test] Error 1
```
It's so not make sense. We should fix the version that we use instead of relying heavily on the latest Pull Request and upgrade the version manually when we need it.

None.

None.

### DIFF
--- a/hack/install/install_cni.sh
+++ b/hack/install/install_cni.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+CNI_VERSION=v0.7
+
 # keep the first one only
 GOPATH="${GOPATH%%:*}"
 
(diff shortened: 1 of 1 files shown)


chuanchang:
@Starnop IMO, it had better to keep consistent with upstream,  you know the cni plugin will constantly change in upstream, so I think it's not a good idea to give fixed version, what do you think? 

BTW, I have an another PR 2452 to fix it.

hack/install/install_cni.sh
```diff
@@ -20,7 +22,14 @@ cni::install_cni() {
   workdir="${GOPATH}/src/${pkg}"
 
   # downloads github.com/containernetworking/plugins
-  go get -u -d "${pkg}"/...
+  if [ ! -d "${workdir}" ]; then
```
mathspanda: If fix cni version, we can just download releases from https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-amd64-${CNI_VERSION}.tgz

Starnop: In fact, the size of the source code package is smaller than the size of the `.tgz`. In addition, we will be freedom to operate the branch . So no need to change it, IMO. WDYT? Anyway, thanks for your advice. :)

fuweid:
> @Starnop IMO, it had better to keep consistent with upstream, you know the cni plugin will constantly change, so I think it's not a good idea to give fixed version, what do you think?

I think we should use the stable version here because the change maybe unstable in master.

PR #2453, opened 2018-11-08
bugfix: maintain a separate id which one-by-one with podSandboxID

Starnop:

As title described, why we maintain a separate id of sandboxMeta?
If we use the id of `sandboxContainer` , we must store the id of sandbox after `sandboxContainer` has been created, if the process exits between creating container completely and storing it, the container will not be controlled by sandboxstore.

So we define the SandboxMeta struct as follows:
```
// SandboxMeta represents the sandbox's meta data.
type SandboxMeta struct {
	// ID is the id of sandbox.
	ID string

	// SandboxContainerID is the container id of sandbox.
	SandboxContainerID string
        ......
}
```
The field `ID` is used for :
1. as the key of sandboxMeta
2. reponse to the caller like kubelet
3. as the value of label `sandboxIDLabelKey = "io.kubernetes.sandbox.id"` of the container 

And the field `SandboxContainerID ` is used for :
1. operate the container of sandbox
2. as the part of `sandboxRootDir` 

None.

None.

### DIFF
--- a/cri/v1alpha2/cri.go
+++ b/cri/v1alpha2/cri.go
@@ -231,6 +231,7 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		metrics.PodActionsTimer.WithLabelValues(label).Observe(time.Since(start).Seconds())
 	}(time.Now())
 
+	removeContainerErr := false
 	config := r.GetConfig()
 
 	// Step 1: Prepare image for the sandbox.
@@ -242,8 +243,43 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 		return nil, err
 	}
 
+	// generate the id of sandbox meta and store it.
+	podSandboxID, err := c.generateSandboxID()
+	if err != nil {
+		return nil, err
+	}
+	sandboxMeta := &SandboxMeta{
+		ID: podSandboxID,
+	}
+	if err := c.SandboxStore.Put(sandboxMeta); err != nil {
+		return nil, err
+	}
+
+	// If running sandbox failed, clean up the sandboxMeta from sandboxStore.
+	defer func() {
+		if retErr != nil && !removeContainerErr {
+			// should not remove the sandbox container metadata from sandboxStore
+			// until it was removed by pouchd.
+			if err := c.SandboxStore.Remove(podSandboxID); err != nil {
+				logrus.Errorf("failed to remove the metadata of container %q from sandboxStore: %v", podSandboxID, err)
+			}
+		}
+	}()
+
+	if _, ok := config.Annotations[anno.LxcfsEnabled]; ok {
+		enableLxcfs, err := strconv.ParseBool(config.Annotations[anno.LxcfsEnabled])
+		if err != nil {
+			return nil, err
+		}
+		sandboxMeta.LxcfsEnabled = enableLxcfs
+
+		if err := c.SandboxStore.Put(sandboxMeta); err != nil {
+			return nil, err
+		}
+	}
+
 	// Step 2: Create the sandbox container.
-	createConfig, err := makeSandboxPouchConfig(config, image)
+	createConfig, err := makeSandboxPouchConfig(config, image, podSandboxID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to make sandbox pouch config for pod %q: %v", config.Metadata.Name, err)
 	}
@@ -254,23 +290,12 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
 	if err != nil {
 		return nil, fmt.Errorf("failed to create a sandbox for pod %q: %v", config.Metadata.Name, err)
 	}
-	id := createResp.ID
+	sandboxContainerID := createResp.ID
 
 	// once sandbox container created, we are obligated to store it.
-	sandboxMeta := &SandboxMeta{
-		ID:      id,
-		Config:  config,
-		Runtime: config.Annotations[anno.KubernetesRuntime],
-	}
-
-	if _, ok := config.Annotations[anno.LxcfsEnabled]; ok {
-		enableLxcfs, err := strconv.ParseBool(config.Annotations[anno.LxcfsEnabled])
-		if err != nil {
-			return nil, err
-		}
-		sandboxMeta.LxcfsEnabled = enableLxcfs
-	}
-
+	sandboxMeta.SandboxContainerID = sandboxContainerID
+	sandboxMeta.Config = config
+	sandboxMeta.Runtime = config.Annotations[anno.KubernetesRuntime]
 	if err := c.SandboxStore.Put(sandboxMeta); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


cri/v1alpha2/cri.go
```diff
@@ -475,6 +499,14 @@ func (c *CriManager) RemovePodSandbox(ctx context.Context, r *runtime.RemovePodS
 
 	podSandboxID := r.GetPodSandboxId()
 
+	// get the sandbox's meta data.
+	sandboxMeta, err := c.getSandboxMeta(podSandboxID)
+	if err != nil {
+		return nil, err
+	}
+
```
fengzixu: It seems like that we should delete this bank line.

cri/v1alpha2/cri_utils.go
```diff
@@ -48,6 +50,36 @@ func toCriTimestamp(t string) (int64, error) {
 	return result.UnixNano(), nil
 }
 
+// generateSandboxID generates an ID for newly created sandbox meta.
+// We must ensure that this ID has not used yet.
+func (c *CriManager) generateSandboxID() (string, error) {
```
fengzixu: Whether we should a ut for this function?

cri/v1alpha2/cri.go
```diff
@@ -964,11 +1002,13 @@ func (c *CriManager) ContainerStatus(ctx context.Context, r *runtime.ContainerSt
 	}
 
 	podSandboxID := container.Config.Labels[sandboxIDLabelKey]
-	res, err := c.SandboxStore.Get(podSandboxID)
+
+	// get the sandbox's meta data.
```
fengzixu: I found that some comments' first alphabet is upper, while others are lower. Should we use the upper style in comment?

cri/v1alpha2/cri.go
```diff
@@ -406,57 +430,57 @@ func (c *CriManager) StopPodSandbox(ctx context.Context, r *runtime.StopPodSandb
...
+			ID:           sandboxContainerID,
 			NetNS:        netNSPath,
 			PortMappings: toCNIPortMappings(sandboxMeta.Config.GetPortMappings()),
 		})
 		if err != nil {
 			if !os.IsNotExist(err) {
 				return nil, err
 			}
-			logrus.Warnf("failed to find network namespace file %s of sandbox %s which may have been already stopped", netNSPath, podSandboxID)
+			logrus.Warnf("failed to find network namespace file %s of sandbox %s which may have been already stopped", netNSPath, sandboxContainerID)
```
fengzixu: Hmmm, I found we use the level of logging without some standards in the pouch. What are the situations we use the `warn` to represent errors? What are the differences between `warn` and `Error`?

fengzixu: Is there any best practice about using log level? @allencloud

cri/v1alpha2/cri.go
```diff
@@ -248,65 +249,83 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
...
+	createResp, err := c.ContainerMgr.Create(ctx, sandboxName, createConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create a sandbox for pod %q: %v", config.Metadata.Name, err)
 	}
+	sandboxContainerID := createResp.ID
 
+	// once sandbox container created, we are obligated to store it.
+	sandboxMeta.SandboxContainerID = sandboxContainerID
+	sandboxMeta.Config = config
+	sandboxMeta.Runtime = config.Annotations[anno.KubernetesRuntime]
```
HusterWan: Just asking, Could the `Annotations` be nil?

cri/v1alpha2/cri.go
```diff
@@ -248,65 +249,83 @@ func (c *CriManager) RunPodSandbox(ctx context.Context, r *runtime.RunPodSandbox
...
 		if retErr != nil {
-			if err := c.ContainerMgr.Remove(ctx, id, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true}); err != nil {
-				logrus.Errorf("failed to remove container when running sandbox failed %q: %v", id, err)
-				return
-			}
-			// should not remove the sandbox container metadata from sandboxStore
-			// until it was removed by pouchd.
-			if err := c.SandboxStore.Remove(id); err != nil {
-				logrus.Errorf("failed to remove the metadata of container %q from sandboxStore: %v", id, err)
+			if err := c.ContainerMgr.Remove(ctx, sandboxContainerID, &apitypes.ContainerRemoveOptions{Volumes: true, Force: true}); err != nil {
```
HusterWan: Can we add some test cases for this situation?

cri/v1alpha2/cri.go
Starnop: I think we should warn level here, if cannot find network namespace file, we should continue to stop the container and give a warning as a hint. 

I have read your comment  https://github.com/alibaba/pouch/issues/2149#issuecomment-430535398 :)

And this are some of my opinions:
I think the error should not happen or seldom. If it happens, the developer should be informed to deal with it;Warning is used to deal with events that have a high probability of occurring, and we've done something to deal with that, just as a notification to you  "hey, look, it's happening again".

Just a few personal points of view，welcome to have a communication.:)

cri/v1alpha2/cri.go
Starnop: In general, we try to use lower case for comments in code snippets.

cri/v1alpha2/cri.go
Starnop: Yeah, and pouchd will check that.

cri/v1alpha2/cri.go
fengzixu: Whatever `warn` or `error` is used to stand for the exception happened in pouch. As you said:
1. `warn` stands for the exception which can be handled(not guarantee success)
2. `error` stands for the exception which cannot be handled.

But, if we repeatedly handle an exception, it always cannot be resolved. Whether this message should be notified to developers? It seems like that `error` and `warn` cannot notify developers efficiently and fastly. So there is not a good reason to use `warn` to instead of `error`.

cri/v1alpha2/cri.go
Starnop: Yeah， I have a test locally:
1. return err directly between success to create sandbox  and start
2. return err directly when remove container for pouchd
3. runpodsandbox
```
# crictl runp pod-config.yaml 
DEBU[0000] RunPodSandboxRequest: &RunPodSandboxRequest{Config:&PodSandboxConfig{Metadata:&PodSandboxMetadata{Name:test-remove-sandbox,Uid:hdishd83dddidwnduwk28bcsb,Namespace:default,Attempt:1,},Hostname:,LogDirectory:,DnsConfig:nil,PortMappings:[],Labels:map[string]string{},Annotations:map[string]string{},Linux:nil,},} 
DEBU[0000] RunPodSandboxResponse: nil                   
FATA[0000] run pod sandbox failed: rpc error: code = Unknown desc = test return error when run PodSandbox
```
4. pouch ps -a
```
# pouch ps -a
Name                                                              ID       Status         Created         Image                                                                 Runtime
k8s_POD_test-remove-sandbox_default_hdishd83dddidwnduwk28bcsb_1   03e587   created        2 minutes ago   registry.cn-hangzhou.aliyuncs.com/google-containers/pause-amd64:3.0   runc
```
5. pouch inspect
```
# pouch inspect 03e587
[
    ......          
            "Labels": {
                "io.kubernetes.pouch.type": "sandbox"
    ......      
]
```
6. ListPodSandbox
```
# crictl pods
DEBU[0000] ListPodSandboxRequest: &ListPodSandboxRequest{Filter:&PodSandboxFilter{Id:,State:nil,LabelSelector:map[string]string{},},} 
DEBU[0000] ListPodSandboxResponse: &ListPodSandboxResponse{Items:[&PodSandbox{Id:03e5874bc39ce08f70f260c00617f81d14242bf39cf53075022071d1af0a66bf,Metadata:&PodSandboxMetadata{Name:test-remove-sandbox,Uid:hdishd83dddidwnduwk28bcsb,Namespace:default,Attempt:1,},State:SANDBOX_NOTREADY,CreatedAt:1541681197387544250,Labels:map[string]string{},Annotations:map[string]string{},}],} 
POD ID              CREATED             STATE               NAME                  NAMESPACE           ATTEMPT
03e5874bc39ce       5 minutes ago       NotReady            test-remove-sandbox   default             1
```

Everything looks good. ^_^

cri/v1alpha2/cri.go
HusterWan: Does CRI check that?

cri/v1alpha2/cri.go
xiaoxubeii: I almost agree with @fengzixu about log definition of warn and error. But in this case: 
```
logrus.Warnf("failed to find network namespace file %s of sandbox %s which may have been already stopped", netNSPath, podSandboxID)
```
It probably likes that it causes application oddities because of not finding network namespace, but maybe automatically recovery which means already stopped. So I prefer to using warn here.

HusterWan:
@Starnop what this PR status? could we make a process?

Starnop:
A separate id is not expected by CRI, so we use the generate id to create the container. refer to ##2503

PR #2460, opened 2018-11-09
improve efficiency to avoid unnecessary system calls

### LINKED ISSUE #2459: Duplicate and unnecessary system call in pkg/kernel/GetVersionInfo()
### Ⅰ. Issue Description

as the image (perf record to flame graph) below, i found a CPU hot point in the pkg/kernel/GetVersionInfo().

this method may only need to be called once, because the kernel version is fixed after the system is up. ?

pouch-kernel-version

1.
2.
3.

### Ⅵ. Environment:

- pouch version (use `pouch version`): 1.0.0
- OS (e.g. from /etc/os-release): CentOS 7.x
- Kernel (e.g. `uname -a`):
- Install tools: rpm -ivh
- Others:

eryx:

cache the kernel version-info and improve syscall efficiency (in some special cases, reduce CPU load about 12%)

fixes #2459

### DIFF
--- a/pkg/kernel/kernel.go
+++ b/pkg/kernel/kernel.go
@@ -20,27 +20,37 @@ func (k *VersionInfo) String() string {
 	return fmt.Sprintf("%d.%d.%d-%s", k.Kernel, k.Major, k.Minor, k.Flavor)
 }
 
+var (
+	kernelVersionCache *VersionInfo
+)
+
 // GetKernelVersion returns the kernel version info.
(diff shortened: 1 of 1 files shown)


eryx:
kexec case!

houstar:
@eryx   Maybe you can add cache-flush time like that,

`cache-flush-time = 5 min`

I encourage you to freely talk to others and make your mind matter 👍

PR #2463, opened 2018-11-10
refactor: use raw system call instead of clone-and-read-from-stdout to boost GetKernelVersion.

SimonCqk:

refactor: use raw system call instead of clone-and-read-from-stdout to boost GetKernelVersion. Original `exec.Run` cloned a new process which caused performance hot spot.

fixes#2459

kernel_test.go has covered this case.


pkg/kernel/kernel.go
```diff
@@ -2,9 +2,7 @@ package kernel
 
 import (
 	"fmt"
```
HusterWan: please add a blank line, thanks.

SimonCqk: sorry neglected that, I've fixed.

SimonCqk:
@fuweid sorry to trouble you, I've done that.

pkg/kernel/kernel.go
```diff
@@ -27,14 +26,15 @@ func GetKernelVersion() (*VersionInfo, error) {
 		flavor               string
 	)
 
-	_, stdout, _, err := exec.Run(0, "uname", "-r")
+	buf := unix.Utsname{}
+	err := unix.Uname(&buf)
```
HusterWan: Can we add a benchmark test for the `Uname` interface?

SimonCqk: Done, plz check it.

pkg/kernel/kernel_test.go
```diff
@@ -3,7 +3,9 @@ package kernel
 import (
 	"testing"
 
+	"github.com/alibaba/pouch/pkg/exec"
```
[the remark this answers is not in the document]
SimonCqk: done.

pkg/kernel/kernel_test.go
```diff
@@ -12,3 +15,23 @@ func TestGetKernelVersion(t *testing.T) {
 
 	println(version.String())
 }
+
+// Benchmark result for below two methods to execute `uname` command in GetKernelVersion().
+
+// BenchmarkGetUnameByUnix-4      	  200000	     10584 ns/op
+// BenchmarkGetUnameByExecRun-4   	     200	   6255530 ns/op
+
+// Benchmark for executing `uname` by raw unix system call
```
HusterWan: Just asking, how long this two Benchmark case execute?

SimonCqk: image
the result shows 4.135s

Issue #2405, opened 2018-11-01
support a common filter for container, volume and images choose


Ace-Tang:
## Why you need it?

```
pouch ps --filter
pouch images --filter
pouch volume --filter
```

pouch ps has implements a simple filter package, and pouch images and pouch volume are not support filter command, should get a common filter package for these command in daemon, and support the command in cli side. Optimize on current fileter package or implement a better one is wlecome.

ZYecho:
As a supplement，also have` pouch events -- filter`. And I plan to reuse the filter package same as `events` to implement  pouch images filter. wdyt?

Ace-Tang:
Great ! /cc me when you finish.

ZYecho:
@Ace-Tang  I revisit the pr about `event filters` and `ps filter`, and I think we can build  a more common filter package using current two packages, I will work on it after #2410 merged.

PR #2469, opened 2018-11-14
feature: add volume filter

ZYecho:

Add `pouch volume ls` filter flag support.Support following filter type:
* driver, use driver name to do accurate match
* name, use volume name to do accurate match
* label, use label=key or label=key=value to do volume label filter

#2462 

Add both api test case and cli test case.

wait for CI.

None.

### DIFF
--- a/apis/filters/parse.go
+++ b/apis/filters/parse.go
@@ -187,3 +187,32 @@ func (args Args) Validate(accepted map[string]bool) error {
 func FamiliarMatch(pattern string, ref string) (bool, error) {
 	return path.Match(pattern, ref)
 }
+
+// MatchKVList returns true if all the pairs in sources exist as key=value
+// pairs in the mapping at key, or if there are no values at key.
+func (args Args) MatchKVList(key string, sources map[string]string) bool {
+	fieldValues := args.fields[key]
+
+	// do not filter if there is no filter set or cannot determine filter
+	if len(fieldValues) == 0 {
+		return true
+	}
+
+	if len(sources) == 0 {
+		return false
+	}
+
+	for value := range fieldValues {
+		attrKV := strings.SplitN(value, "=", 2)
+
+		v, ok := sources[attrKV[0]]
+		if !ok {
+			return false
+		}
+		if len(attrKV) == 2 && attrKV[1] != v {
+			return false
+		}
+	}
+
+	return true
+}
--- a/client/volume_list_test.go
+++ b/client/volume_list_test.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/alibaba/pouch/apis/filters"
 	"github.com/alibaba/pouch/apis/types"
 
 	"github.com/stretchr/testify/assert"
@@ -19,7 +20,7 @@ func TestVolumeListServerError(t *testing.T) {
 	client := &APIClient{
 		HTTPCli: newMockClient(errorMockResponse(http.StatusInternalServerError, "Server error")),
 	}
-	_, err := client.VolumeList(context.Background())
+	_, err := client.VolumeList(context.Background(), filters.NewArgs())
 	if err == nil || !strings.Contains(err.Error(), "Server error") {
 		t.Fatalf("expected a Server Error, got %v", err)
 	}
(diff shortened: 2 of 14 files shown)


storage/volume/core.go
```diff
@@ -272,14 +274,25 @@ func (c *Core) ListVolumes(labels map[string]string) ([]*types.Volume, error) {
 		retVolumes = append(retVolumes, v)
 	}
 
-	// filter volumes if specify labels
-	if len(labels) != 0 {
+	// filter volumes
+	if filter.Len() > 0 {
```
HusterWan: How about 
```
if filter.Len() == 0 {
    return retVolumes, nil
}
```

we can decrease the indent, WDYT?

apis/filters/parse.go
```diff
@@ -187,3 +187,32 @@ func (args Args) Validate(accepted map[string]bool) error {
 func FamiliarMatch(pattern string, ref string) (bool, error) {
 	return path.Match(pattern, ref)
 }
+
+// MatchKVList returns true if all the pairs in sources exist as key=value
+// pairs in the mapping at key, or if there are no values at key.
+func (args Args) MatchKVList(key string, sources map[string]string) bool {
```
zhuangqh: This part of code is similar with https://github.com/alibaba/pouch/blob/master/daemon/mgr/container_list.go#L74
Is there any plan to unify them?

client/volume_list_test.go
```diff
@@ -12,14 +12,15 @@ import (
 
 	"github.com/alibaba/pouch/apis/types"
 
+	"github.com/alibaba/pouch/apis/filters"
```
fuweid: we could add the package into second group

apis/filters/parse.go
ZYecho: Yes, there's a plan, need to refactor to support a common filter package, as #2405.

test/cli_volume_test.go
```diff
@@ -327,4 +327,44 @@ func (suite *PouchVolumeSuite) TestVolumeListOptions(c *check.C) {
...
+
+	// test filter options
+	volumeName4 := "volume_" + funcname + "4"
+	command.PouchRun("volume", "create", "--name", volumeName4, "--driver", "local", "--label", "test=foo").Assert(c, icmd.Success)
+	defer command.PouchRun("volume", "rm", volumeName4)
+	// test name filter
+	res := command.PouchRun("volume", "list", "--filter", "name="+volumeName4)
+	res.Assert(c, icmd.Success)
+
+	found := 0
```
fuweid: could we reorder the cases here? for example, use test case table to do that. I think there are a lot of duplicate code here. thanks.

storage/volume/core.go
```diff
@@ -272,29 +274,40 @@ func (c *Core) ListVolumes(labels map[string]string) ([]*types.Volume, error) {
...
-		}
+	// filter volumes
+	if filter.Len() == 0 {
+		return retVolumes, nil
+	}
 
-		return filteredVolumes, nil
+	filteredVolumes := make([]*types.Volume, 0)
+	for _, vol := range retVolumes {
+		// do label filter
```
fuweid: just for confirm, filter only support `and` right?

ZYecho: relationship between different filter was `and`, but if multi same filter was use, such as `--driver=A, --driver=B`, there will be A or B.

fuweid: good to know. Thanks

test/cli_volume_test.go
ZYecho: thanks for remind me, update.

test/cli_volume_test.go
```diff
@@ -350,4 +350,29 @@ func (suite *PouchVolumeSuite) TestVolumeListOptions(c *check.C) {
...
+	// test filter options
+	volumeName4 := "volume_" + funcname + "4"
+	command.PouchRun("volume", "create", "--name", volumeName4, "--driver", "tmpfs", "--label", "test=foo").Assert(c, icmd.Success)
+	defer command.PouchRun("volume", "rm", volumeName4)
+
+	// test name, label, driver filter
+	filterArgs := []string{
+		"name=" + volumeName4,
+		"label=test",
+		"driver=tmpfs",
```
Ace-Tang: how about add some `and` case, like `--lable=test --name=a --driver=tmpfs`

ZYecho: Agree.

HusterWan:
LGTM

PR #2510, opened 2018-11-27
bugfix: cleanup containerd container before do restart policy

fuweid:

If the cleanup action is taken before restart policy, the restart action
will be failed because new container keeps the same name with the old
one. We should cleanup the old one before the restart policy.

In order to make it happen, we add the cleanup function in the hooks.

> NOTE: the cleanup should be done after update status because we should make sure that all the status has been updated in the pouch container store.

NONE

no need

Use the following command and the container should be restarted after exit.

```
sudo pouch run --restart=always -it busybox sh -c "exit 1"
```

And don't break existing case

### DIFF
--- a/ctrd/watch.go
+++ b/ctrd/watch.go
@@ -36,7 +36,7 @@ func (m *Message) ExitTime() time.Time {
 type watch struct {
 	sync.Mutex
 	containers map[string]*containerPack
-	hooks      []func(string, *Message) error
+	hooks      []func(string, *Message, func() error) error
 
 	// containerdDead to specify whether containerd process is dead
 	containerdDead bool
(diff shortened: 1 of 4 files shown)


ctrd/watch.go
```diff
@@ -98,23 +98,31 @@ func (w *watch) add(pack *containerPack) {
...
 		if !pack.skipStopHooks {
 			for _, hook := range w.hooks {
-				if err := hook(pack.id, msg); err != nil {
+				if err := hook(pack.id, msg, cleanupFunc); err != nil {
 					logrus.Errorf("failed to execute the exit hooks: %v", err)
 					break
 				}
 			}
 		}
+		cleanupFunc()
```
HusterWan: `if !pack.skipStopHooks`, the `cleanupFunc` will be called twice?

HusterWan: I seen, you use the `sync.Once` to control the `cleanup` function can only be called once.

I check the `sync.Once` feature source code : 
```
// Once is an object that will perform exactly one action.
type Once struct {
	m    Mutex
	done uint32
}

// Do calls the function f if and only if Do is being called for the
// first time for this instance of Once. In other words, given
// 	var once Once
// if once.Do(f) is called multiple times, only the first call will invoke f,
// even if f has a different value in each invocation. A new instance of
// Once is required for each function to execute.
//
// Do is intended for initialization that must be run exactly once. Since f
// is niladic, it may be necessary to use a function literal to capture the
// arguments to a function to be invoked by Do:
// 	config.once.Do(func() { config.init(filename) })
//
// Because no call to Do returns until the one call to f returns, if f causes
// Do to be called, it will deadlock.
//
// If f panics, Do considers it to have returned; future calls of Do return
// without calling f.
//
func (o *Once) Do(f func()) {
	if atomic.LoadUint32(&o.done) == 1 {
		return
	}
	// Slow-path.
	o.m.Lock()
	defer o.m.Unlock()
	if o.done == 0 {
		defer atomic.StoreUint32(&o.done, 1)
		f()
	}
}
```

This function just can be called only once,  am I right?

zhuangqh: The function `cleanupOnce.Do` could be called multi times. But the function `f` as the parameter of `Do` would be exactly called once. @HusterWan

HusterWan: @zhuangqh I got you!

HusterWan:
LGTM

PR #2518, opened 2018-11-30
WIP feature: implement a stream mechanism to abstract std io

### LINKED ISSUE #2490: [feature request] implement a stream mechanism to abstract stdin, stdout and stderr in cmd side
## Why you need it?

We should implement a unified stream mechanism for standard io, such as stdin, stdout and stderr by cmd side. And the mechanism would handle copying input to and output from to the connection and also be reused in where we need attach or access container io.

## How it could be?

Code snippet is something like that:

```
type containerIOStreamer struct {
	inputStream  io.ReadCloser
	outputStream io.Writer
	errorStream  io.Writer

	conn net.Conn

	tty        bool
	detachKeys string
}
```

We can begin to proxy container io by:

```
streamer := containerIOStreamer{
	inputStream:  os.Stdin,
	outputStream: os.Stdout,
	errorStream:  os.Stderr,
	conn:         conn,
	tty:          config.Tty,
	detachKeys:   options.DetachKeys,
}
streamer.stream(ctx)
```
## Other related information

NONE.

xiaoxubeii:
It implements a unified stream mechanism for standard io, such as stdin, stdout and stderr on cli side. And it also implements detachkeys on cli/run to verify the correctness of stream mechanism. 

fixes #2490, #2221

```
$ pouch run -it --detach-keys="ctrl-x" busybox /bin/sh
```
It would close container tty when type ctrl-x.

### DIFF
--- a/pkg/ioutils/hijack.go
+++ b/pkg/ioutils/hijack.go
@@ -1,0 +1,151 @@
+package ioutils
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"sync"
+
+	"github.com/docker/docker/pkg/ioutils"
+	"github.com/docker/docker/pkg/term"
+	"github.com/sirupsen/logrus"
+)
+
+var defaultEscapeKeys = []byte{16, 17}
+
+// HijackedIOStreamer represents io stream
+type HijackedIOStreamer struct {
+	Streams      Streams
+	InputStream  io.ReadCloser
+	OutputStream io.Writer
+
+	Conn      net.Conn
+	BufReader *bufio.Reader
+
+	Tty        bool
+	DetachKeys string
+}
+
+// Stream is used to streaming io
+func (h *HijackedIOStreamer) Stream(ctx context.Context) error {
+	restoreInput, err := h.setupInput()
+	if err != nil {
+		return fmt.Errorf("unable to setup input stream: %s", err)
+	}
+
+	defer restoreInput()
+
+	outputDone := h.beginOutputStream(restoreInput)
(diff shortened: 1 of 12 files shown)


pkg/ioutils/hijack.go
```diff
@@ -0,0 +1,232 @@
...
+	// As the client exits right after, simply discard the close
+	// until we find a better solution.
+	//
+	// This can also cause the client on Windows to get stuck in Win32 CloseHandle()
+	// in some cases. See https://github.com/docker/docker/issues/28267#issuecomment-288237442
+	// Tracked internally at Microsoft by VSO #11352156. In the
+	// Windows case, you hit this if you are using the native/v2 console,
+	// not the "legacy" console, and you start the client in a new window. eg
+	// `start docker run --rm -it microsoft/nanoserver cmd /s /c echo foobar`
+	// will hang. Remove start, and it won't repro.
```
fuweid: to be honest, I don't think we can copy the code and put it into pouch code base.
Welcome to use vendor or write it by yourself.

Thanks

Ace-Tang: Agree, @xiaoxubeii actually, pouch's io stream not same with moby, so that code not suit for pouch

xiaoxubeii: @Ace-Tang Actually I have changed some docker codes to be compatible with pouch and it has worked. But after discussing with fuweid, there are maybe some problems to copy docker codes to pouch codebase directly. So I am working on rewriting it.

xiaoxubeii:
@fuweid Request for review : )

PR #2523, opened 2018-12-02
cleanup: rm runtime.Caller with const func name

ZYecho:

As the title desc, just clean up some test code to make more readable.

None.

None.

wait for CI pass.

None.


fuweid:
@ZYecho I think we can upgrade the go-check package so that we can use https://godoc.org/gopkg.in/check.v1#C.TestName to get the function name.

PR #2530, opened 2018-12-04
[WIP] Upgrade to containerd v1.2

fuweid:

The containerd has been improved a lot in v1.2, which makes the service
stable than before. Based on this, we need to do the upgrade.

```
using following commands to update the vendor

govendor fetch github.com/containerd/containerd/...@v1.2.0
govendor fetch github.com/gogo/protobuf/...@v1.0.0
govendor fetch github.com/gogo/googleapis/...@08a7655d27152912db7aaf4f983275eaf8d128ef
govendor fetch google.golang.org/grpc/...@v1.12.0
govendor fetch github.com/golang/protobuf/...@v1.1.0
```

Basically, both the contaienrd and shim API interfaces are stable so
that we don't need to worry about the remote API calls. However, the
package level interface has been changed a lot. The commit is used to
align with containerd@v1.2.0.

    1. events part:
       using exchange event helper instead of raw grpc conn

    2. image part:
      a. import:
         new Import interface makes the import easier without Importer.

      b. commit:
         content helper has been upgraded with less arguments for caller.
         and `diff` package separates the Applier and Comparer interfaces.
         The changes can make the code clear in PouchContainer

    3. lease:
      containerd client doesn't provide ListLeases interface directly.
      PouchContainer needs to get LeasesService and make the call by itself.

    4. container IO:
      remove codes which comes from containerd. The code was used to align
      with containerd v1.2, which can make upgrade work easier. For now, we
      can remove this.

I also use subshell to run the `integration::ping_pouchd` command so that the error log can show up in CI.

NONE

no need

use existing CI testing case

I separate the change into two commits because the vendor change is too big. 
There are somethings need to do after this PR, for example, it is time to remove the ctrd/wrapper_client.go because the grpc has been updated and it can support the [defaultMaxStreamsClient](https://github.com/grpc/grpc-go/blob/11b582728a13ef54e14e4fa930d693de2e32420f/internal/transport/defaults.go#L34)

related to https://github.com/containerd/containerd/issues/2862


fuweid:
the first commit is for vendor update and the second one is for code change. hope it can help.

fuweid:
still investigate the failed cases about load/save functionality. It seems that the containerd export functionality doesn't support the image which has several manifests.

Ace-Tang:
Hold this pr

fuweid:
close it and re-create other one

PR #2536, opened 2018-12-05
fix: compatible with multi host ip

zhuangqh:

make PouchContainer compatible with multi host ip.

result returned by `hostname -i` is split by space, but scanner split tokens by lines by default.

### DIFF
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -119,6 +119,7 @@ func GetNodeIP() string {
 	}
 
 	scanner := bufio.NewScanner(bytes.NewReader(output))
+	scanner.Split(bufio.ScanWords)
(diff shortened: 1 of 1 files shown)


pkg/system/system.go
```diff
@@ -119,6 +119,7 @@ func GetNodeIP() string {
 	}
 
 	scanner := bufio.NewScanner(bytes.NewReader(output))
+	scanner.Split(bufio.ScanWords)
```
HusterWan: can we add some comments here?

rudyfly: @zhuangqh PTAL

rudyfly:
LGTM

PR #2547, opened 2018-12-10
test: update the existing tag image case

fuweid:

The TestImageTagFailToOverrideExistingTag tests the case that it will
fail to use existing tag name to override existing tag.

NONE

update existing case

wait for CI

### DIFF
--- a/test/api_image_tag_test.go
+++ b/test/api_image_tag_test.go
@@ -30,7 +30,7 @@ func (suite *APIImageTagSuite) SetUpTest(c *check.C) {
 // TestImageTagCreateWithTagOK tests OK.
 func (suite *APIImageTagSuite) TestImageTagCreateWithTagOK(c *check.C) {
 	repo, tag := "localhost:5000/testimagetagok/pouch", "0.5.0"
-	resp, err := suite.sendRequest(busyboxImage, repo, tag)
+	resp, err := suite.doTag(busyboxImage, repo, tag)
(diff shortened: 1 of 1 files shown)


test/api_image_tag_test.go
```diff
@@ -73,14 +73,29 @@ func (suite *APIImageTagSuite) TestImageTagFailToOverrideExistingPrimaryReferenc
...
-	c.Assert(err, check.IsNil)
-	CheckRespStatus(c, resp, 400)
+	// create valid tag
+	{
+		command.PouchRun("pull", busyboxImage).Assert(c, icmd.Success)
+
+		resp, err := suite.sendRequest(busyboxImage, repo, tag)
+		c.Assert(err, check.IsNil)
+		CheckRespStatus(c, resp, 201)
+		defer DelImageForceOk(c, tagRef)
```
rudyfly: why delete tagRef image？

fuweid: the `tagRef` image is created by the case. If we don't delete it, re-run this case will fail. It just is used to clean up.

rudyfly: But the `busyboxImage` also will be remain, dont need to delele it？

fuweid: `busyboxImage` is created for the case suite and basically, all suites are related to this image. Therefore, we don't need to delete it.

rudyfly:
LGTM

PR #2549, opened 2018-12-11
Get  a valid image user and image uid

### LINKED ISSUE #2548: [Bug] Cannot get a valid user name by CRI ImageStatus or ListImages interface
### Ⅰ. Issue Description
When defines a container runtime user name in image, e.g.
image

Kubelet will read and parse it from CRI ImageStatus interface , and pass runtime user to createContainer CRI. 
In kubelet logic , first choose use uid if it is not nil, then we choose user name if it is not empty,
image

But in pouch CRI logic, ImageStatus and ListImages interface call `imageToCriImage`, which it will initialize `Uid` as `runtime.Int64Value{}`, so in any case, the `Uid` will not be nil
image
So if user define user in image, it can't work as user expect.

Return real uid and username in image info

### Ⅵ. Envronment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

changyaowei:
Get a valid user and image uid through imageStatus and ListImages CRI 

fixes #2548

Modify an exist test case

### DIFF
--- a/cri/v1alpha2/cri_utils.go
+++ b/cri/v1alpha2/cri_utils.go
@@ -892,23 +892,25 @@ func containerNetns(container *mgr.Container) string {
 
 // imageToCriImage converts pouch image API to CRI image API.
 func imageToCriImage(image *apitypes.ImageInfo) (*runtime.Image, error) {
-	uid := &runtime.Int64Value{}
-	imageUID, username := getUserFromImageUser(image.Config.User)
-	if imageUID != nil {
-		uid.Value = *imageUID
+	if image == nil || image.Config == nil {
+		return nil, fmt.Errorf("unable to convert a nil pointer to a runtime API image")
 	}
-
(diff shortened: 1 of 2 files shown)


cri/v1alpha2/cri_utils.go
```diff
@@ -892,23 +892,25 @@ func containerNetns(container *mgr.Container) string {
 
 // imageToCriImage converts pouch image API to CRI image API.
 func imageToCriImage(image *apitypes.ImageInfo) (*runtime.Image, error) {
-	uid := &runtime.Int64Value{}
-	imageUID, username := getUserFromImageUser(image.Config.User)
-	if imageUID != nil {
-		uid.Value = *imageUID
+	if image == nil || image.Config == nil {
```
HusterWan: Good catch

HusterWan:
LGTM

PR #2554, opened 2018-12-12
bugfix: can't copy data with dr mode

rudyfly:
use volume with dr mode, can't copy image's data to parent directory
when sub-directory has copied. It causes the volume directory is not
empty, so before copy image data, it need to sort mountpoint by the
string length of destination directory.

NONE

add ut: `TestSortMountPoint`
add integration test: `TestRunCopyDataWithDR`

run test `TestRunCopyDataWithDR` pass

### DIFF
--- a/daemon/mgr/container_storage_test.go
+++ b/daemon/mgr/container_storage_test.go
@@ -1,0 +1,71 @@
+package mgr
+
+import (
+	"testing"
+
+	"github.com/alibaba/pouch/apis/types"
+)
+
+func TestSortMountPoint(t *testing.T) {
+	mounts := []*types.MountPoint{
+		{
+			Destination: "/home/admin/log",
+			Source:      "/pouch/volume1",
+		},
+		{
+			Destination: "/var/lib",
+			Source:      "/pouch/volume2",
+		},
+		{
+			Destination: "/home/admin",
+			Source:      "/pouch/volume3",
+		},
+		{
+			Destination: "/home",
+			Source:      "/pouch/volume4",
+		},
+		{
+			Destination: "/opt",
+			Source:      "/pouch/volume5",
+		},
+		{
+			Destination: "/var/log",
+			Source:      "/pouch/volume6",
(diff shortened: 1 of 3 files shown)


fuweid:
link to #2200

daemon/mgr/container_storage_test.go
```diff
@@ -0,0 +1,56 @@
...
+			Source:      "/pouch/volume5",
+		},
+		{
+			Destination: "/var/log",
+			Source:      "/pouch/volume6",
+		},
+	}
+
+	mounts = sortMountPoint(mounts)
+	if mounts[0].Destination != "/opt" || mounts[0].Source != "/pouch/volume5" {
```
fuweid: I think we can prepare two arrays here. one is out of order, other one is in order. After `sortMountPoint`, use loop to check each item. WDYT?

test/cli_run_volume_test.go
```diff
@@ -395,3 +395,30 @@ func (suite *PouchRunVolumeSuite) TestRunWithDiskQuota(c *check.C) {
...
+	res.Assert(c, icmd.Success)
+	if !strings.Contains(res.Stdout(), "spool") ||
+		!strings.Contains(res.Stdout(), "www") {
+		c.Fatal("no copy image data, spool and www")
+	}
+
+	res = command.PouchRun("exec", cname, "ls", "/var/spool")
+	res.Assert(c, icmd.Success)
+	if !strings.Contains(res.Stdout(), "mail") {
+		c.Fatal()
```
fuweid: fatal for what?

rudyfly: 😆 I forgot.

daemon/mgr/container_storage_test.go
rudyfly: Agree.

rudyfly:
@fuweid I have modified test case, PTAL again. 🍵 ☕️

PR #2557, opened 2018-12-13
bugfix: do not use string to convert int64

fuweid:

We should use imageSize.String() instead of string(imageSize).
Otherwise, we will get messy display.

NONE

No need.

before fix:

```
➜  pouch git:(bugfix_image_list_display_issue) sudo pouch images
59788edf1f3e   registry.hub.docker.com/library/busybox:latest   򲸜
```
after fix:

```
➜  pouch git:(bugfix_image_list_display_issue) sudo pouch images
59788edf1f3e   registry.hub.docker.com/library/busybox:latest   715.53 KB
```

### DIFF
--- a/cli/image_list.go
+++ b/cli/image_list.go
@@ -108,9 +108,9 @@ func (i *ImagesCommand) runImages(args []string) error {
 
 	for _, dimg := range dimgs {
 		if i.flagDigest {
-			display.AddRow([]string{dimg.id, dimg.name, dimg.digest, string(dimg.size)})
+			display.AddRow([]string{dimg.id, dimg.name, dimg.digest, dimg.size.String()})
 		} else {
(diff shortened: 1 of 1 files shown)


fengzixu:
It is so interesting problem. :)

fuweid:
I was thinking that it is bug in golang.

ZYecho:
LGTM wait for CI pass

zhuangqh:
LGTM

zhuangqh:
> Converting a signed or unsigned integer value to a string type yields a string containing the UTF-8 representation of the integer. Values outside the range of valid Unicode code points are converted to "\uFFFD".

https://golang.org/ref/spec#Conversions  search subtitle "Conversions to and from a string type"

PR #2565, opened 2018-12-17
test: http_utils_test: add cases for func BoolValue

sanmaopep:
add unit test case for func BoolValue

NONE

go test

### DIFF
--- a/pkg/httputils/http_utils_test.go
+++ b/pkg/httputils/http_utils_test.go
@@ -2,25 +2,48 @@ package httputils
 
 import (
 	"net/http"
+	"strings"
 	"testing"
 )
 
 func TestBoolValue(t *testing.T) {
-	type args struct {
-		r *http.Request
-		k string
-	}
 	tests := []struct {
-		name string
-		args args
-		want bool
+		name    string
+		testStr string
+		want    bool
 	}{
-		// TODO: Add test cases.
+		{"test normal string", "normal string", true},
+		{"test ''", "", false},
+		{"test 0", "0", false},
+		{"test no", "no", false},
+		{"test NO", "NO", false},
+		{"test false", "false", false},
+		{"test FALSE", "FALSE", false},
+		{"test none", "none", false},
+		{"test NONE", "NONE", false},
+		// test trim
+		{"test '  '", "  ", false},
+		{"test trim 0", "  0  ", false},
+		{"test trim no", "  no  ", false},
+		{"test trim NO", "  NO ", false},
+		{"test trim false", " false ", false},
+		{"test trim FALSE", "  FALSE ", false},
+		{"test trim none", "  none ", false},
+		{"test trim NONE", "  NONE  ", false},
 	}
+	boolName := "name"
 	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


pkg/httputils/http_utils_test.go
```diff
@@ -2,24 +2,52 @@ package httputils
...
+		name    string
+		testStr string
+		want    bool
 	}{
-		// TODO: Add test cases.
+		{"test normal string", "normal string", true},
+		{"test ''", "", false},
+		{"test 0", "0", false},
+		{"test no", "no", false},
+		{"test No", "No", false},
```
ZYecho: duplicate?

pkg/httputils/http_utils_test.go
```diff
@@ -2,24 +2,52 @@ package httputils
...
+		{"test No", "No", false},
+		{"test NO", "NO", false},
+		{"test false", "false", false},
+		{"test False", "False", false},
+		{"test FALSE", "FALSE", false},
+		{"test none", "none", false},
```
ZYecho: about the capital test case, I think the NONE and none is enough, no need to add a None or noNe, WDYT?

pkg/httputils/http_utils_test.go
```diff
@@ -2,24 +2,52 @@ package httputils
...
+		{"test trim FALSE", "  FALSE ", false},
+		{"test trim none", "  none ", false},
+		{"test trim None", "  None  ", false},
+		{"test trim NONE", "  NONE  ", false},
 	}
 	for _, tt := range tests {
+		request, _ := http.NewRequest(
+			"POST",
+			"http://test",
+			strings.NewReader("name="+tt.testStr))
```
ZYecho: I think define a var, boolName = "name" is better, because you also use it in the above code, just in case they're divergent.

ZYecho:
@sanmaopep  thanks for your contribution, just a little suggestions.

sanmaopep:
@ZYecho I will modify them and appreciate your suggestions for my first PR 😄

pkg/httputils/http_utils_test.go
```diff
@@ -2,24 +2,46 @@ package httputils
...
+			"POST",
+			"http://test",
+			strings.NewReader(boolName+"="+tt.testStr))
+
+		request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
 		t.Run(tt.name, func(t *testing.T) {
-			if got := BoolValue(tt.args.r, tt.args.k); got != tt.want {
+			if got := BoolValue(request, boolName); got != tt.want {
 				t.Errorf("BoolValue() = %v, want %v", got, tt.want)
```
ZYecho: maybe we could print tt.name to help us find which testcase is wrong?

sanmaopep: it seems that t.Run(tt.name,...) will notify us which testcase is wrong like this:
image

sanmaopep: @ZYecho I've added it

allencloud:
Thanks for your great work. @sanmaopep 
Merging...

PR #2569, opened 2018-12-18
feature: add a paramter labels to CreateSnapshot

firjest:
Signed-off-by: yifang.jy <yifang.jy@alibaba-inc.com>

snapshot interface has labels parameter, so add this parameter to pouch CreateSnapshot In some case use can pass labels.

NO

NO

NO


ctrd/interface.go
```diff
@@ -93,7 +93,7 @@ type ImageAPIClient interface {
 // SnapshotAPIClient provides access to containerd snapshot features
 type SnapshotAPIClient interface {
 	// CreateSnapshot creates a active snapshot with image's name and id.
-	CreateSnapshot(ctx context.Context, id, ref string) error
+	CreateSnapshot(ctx context.Context, id, ref string, labels *map[string]string) error
```
Ace-Tang: Why we should use `*map[string]string`

ctrd/snapshot.go
```diff
@@ -31,8 +31,13 @@ func (c *Client) CreateSnapshot(ctx context.Context, id, ref string) error {
 		return err
 	}
 
+	var opts []snapshots.Opt
+	if labels != nil && len(*labels) > 0 {
+		opts = append(opts, snapshots.WithLabels(*labels))
```
Ace-Tang: How about just `opts := []snapshots.Opt{snapshots.WithLabels(labels)}`, of course, not use `*map` here, see last comment

daemon/mgr/container.go
```diff
@@ -383,7 +383,7 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 
 	snapID := id
 	// create a snapshot with image.
-	if err := mgr.Client.CreateSnapshot(ctx, snapID, config.Image); err != nil {
+	if err := mgr.Client.CreateSnapshot(ctx, snapID, config.Image, nil); err != nil {
```
Ace-Tang: Seems we should get somewhere to pass the label, or it will always be nil

Ace-Tang: We can add a `snapshot-opt` which keep options for snapshotter

ctrd/interface.go
firjest: done

ctrd/snapshot.go
firjest: done

wangforthinker:
To support this feature, containerd should make some changes like this! https://github.com/alibaba/containerd/pull/5/files#diff-491f07c28eb851b720f11ce0d3dbb4f8R331 
. In current version of v1.0.3, containerd do not pass these opts to snapshotter when call method Prepare.

PR #2581, opened 2018-12-19
test: add test cases for mountpoint_test.go and portbindings_test.go

sanmaopep:


chuanchang:
@sanmaopep Need  to fix gofmt issues firstly and then rebase

PR #2608, opened 2018-12-24
feature: split image pull interface

Ace-Tang:
Actually, there is no real Pull interface in containerd, image pull
do not need to unpack but only run a container need. Split Pull
operation into two operations, fetch and unpack. Let pouchd deal with
pulled image.


chuanchang:
LGTM

Ace-Tang:
@fuweid ， I move close json stream to daemon package

PR #2611, opened 2018-12-25
config: disable snapshot syncer by default

fuweid:

The CRI can report the disk usage of contaienr with inode number.
However, the iteration will take long time if the number of files are
too much. Based on this, I think we should disable it by default. If
user wants to use it, set it in config.

NONE

no need

check the log


Ace-Tang:
Will it effect out production ?

fuweid:
> Will it effect our production ?

Not just us, if the container creates a lot of small files and user enables CRI, it will impact the performance and containerd will consume more cpu here.

PR #2621, opened 2018-12-27
test: optimize the time of run test cases

HusterWan:

Since our integration test cases run about 30mins,  the more test cases, the longer the running time.

So, we can optimize some test cases to decrease the running time, like:
  * first stop the container with timeout 1 then delete it
  * add timeout to 1 when stop a container

Let's see if we can decrease the running time of the integration test cases :)


HusterWan:
before this pr:

image


after this pr:

image



**Saved 3 minutes**

allencloud:
This helps a lot for https://github.com/alibaba/pouch/issues/1184.

allencloud:
LGTM

PR #2644, opened 2019-01-02
bugfix: remove default value of cgroups parent

Ace-Tang:
cgroups parent should be empty, since we will give default
cgroups parent in pouch daemon.
```
    cgroupsParent := "default"

    if c.HostConfig.CgroupParent != "" {
        cgroupsParent = c.HostConfig.CgroupParent
    }
```


fuweid:
duplicated with https://github.com/alibaba/pouch/pull/2616

PR #2656, opened 2019-01-09
test: try to make TestRunRestartPolicyNone case happy

HusterWan:

As the TestRunRestartPolicyNone becomes a flaky test case, so we have to try to make it happy


rudyfly:
LGTM

PR #2660, opened 2019-01-11
delete omitempty from Privileged in api

HusterWan:

We have some situations need to check the `Privileged` value in a container meta. So we must make sure the json data of `inspect` api contains the `Privileged` parameter.

I delete the `omitempty` from `Privileged` parameter in `HostConfig` struct 

add the test code in PouchInspectSuite.TestInspectCreateAndStartedFormat

### DIFF
--- a/apis/swagger.yml
+++ b/apis/swagger.yml
@@ -2273,9 +2273,9 @@ definitions:
 
   HostConfig:
     description: "Container configuration that depends on the host we are running on"
-    type: "object"
     allOf:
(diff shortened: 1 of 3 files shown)


HusterWan:
> Do we have other filed in `HostConfig` that inspect need but have `omitempty`, @HusterWan , we need to
> arrange them

I have asked sigma test colleague, so far they just need the `Privileged` parameters.

Maybe we can add more parameters when needed

PR #2665, opened 2019-01-15
bugfix: flaky test case in cli_start_test.go

fuweid:

When we create container, we should have ability to open STDIN even
though we don't attach it yet. Otherwise, the `pouch start` command's
attach with stdin will not work well.

Therefore, `pouch create` need to support `-i` flag to keep STDIN open.
And in order to make the case easy to read, we can use `StdinPipe`
instead of `github.com/kr/pty` which we don't need any more.

NONE

no need.

check it manually and use test case to verify.

if we don't keep stdin open, the `start` will return immediately even though with `-a -i`.
```
# first, create container without openStdin
$ sudo pouch create --name t1 ubuntu cat

# start it
$ pouch start -a -i t1
```

if we keep stdin open, it will act like echo server.

```
➜  test git:(bugfix_container_io) sudo pouch create --name t2 -i ubuntu cat
106adbedb380377a76b814f221fa503bdb6211fbfcc55f68ca8ea2e9d467eeec
➜  test git:(bugfix_container_io) sudo pouch start -a -i t2
skdfj
skdfj
sfkj
sfkj
```


rudyfly:
LGTM, wait ci pass

PR #2668, opened 2019-01-15
bugfix: should initialize ContainerIO for all container

fuweid:

For now, when pouchd restarts, the restore function only re-connects to
existing running/pause containers with initializing container IO.
However, the stopped container also needs the container IO if the user
want to restart the container. For the case, we need to initialize
container IO for all existing container, not just for running/pause
containers.

none

add one

CI

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -256,19 +256,19 @@ func (mgr *ContainerManager) Restore(ctx context.Context) error {
 	containers, err := mgr.List(ctx,
 		&ContainerListOption{
 			All: true,
-			FilterFunc: func(c *Container) bool {
-				return (c.IsRunning() || c.IsPaused())
-			}})
+		},
+	)
 	if err != nil {
-		logrus.Errorf("failed to get container list when restore alive containers: %v", err)
+		logrus.Errorf("failed to get container list when restore containers: %v", err)
 		return errors.Wrap(err, "failed to get container list")
 	}
 
-	// start recover all alive containers
 	for _, c := range containers {
-		logrus.Debugf("Start recover container %s", c.Key())
 		id := c.Key()
-		// recover the running or paused container.
+
+		// NOTE: when pouch is restarting, we need to initialize
+		// container IO for the existing containers just in case that
+		// user tries to restart the stopped containers.
 		cntrio, err := mgr.initContainerIO(c)
 		if err != nil {
 			logrus.Errorf("failed to init container IO %s: %v", id, err)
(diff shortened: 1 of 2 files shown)


test/z_cli_daemon_test.go
```diff
@@ -470,6 +470,47 @@ func (suite *PouchDaemonSuite) TestDaemonWithMultiRuntimes(c *check.C) {
 	dcfg2.KillDaemon()
 }
 
+// TestRestartStoppedContainerAfterDaemonRestart is used to test the case that
+// when container is stopped and then pouchd restarts, the restore logic should
+// initialize the existing container IO settings even though they are not alive.
+func (suite *PouchDaemonSuite) TestRestartStoppedContainerAfterDaemonRestart(c *check.C) {
+	cfg := daemon.NewConfig()
```
rudyfly: Is start daemon with config file? or with command line as before?
And we need think about alios test.

daemon/mgr/container.go
```diff
@@ -256,19 +256,20 @@ func (mgr *ContainerManager) Restore(ctx context.Context) error {
...
 		return errors.Wrap(err, "failed to get container list")
 	}
 
 	// start recover all alive containers
 	for _, c := range containers {
-		logrus.Debugf("Start recover container %s", c.Key())
 		id := c.Key()
-		// recover the running or paused container.
+
+		// NOTE: when pouch restarts, we need to initialize container IO
```
rudyfly: s/restarts/is restarting

daemon/mgr/container.go
```diff
@@ -280,6 +281,13 @@ func (mgr *ContainerManager) Restore(ctx context.Context) error {
 			return err
 		}
 
+		// recover the running or paused container.
+		if !(c.IsRunning() || c.IsPaused()) {
```
rudyfly: running or paused containers don't to recover?

daemon/mgr/container.go
```diff
@@ -256,19 +256,20 @@ func (mgr *ContainerManager) Restore(ctx context.Context) error {
...
-			}})
+		},
+	)
 	if err != nil {
-		logrus.Errorf("failed to get container list when restore alive containers: %v", err)
+		logrus.Errorf("failed to get container list when restore containers: %v", err)
 		return errors.Wrap(err, "failed to get container list")
 	}
 
 	// start recover all alive containers
```
HusterWan: delete this line

daemon/mgr/container.go
fuweid: no need to recover the created/stopped container

daemon/mgr/container.go
fuweid: will do.

test/z_cli_daemon_test.go
fuweid: synced with @rudyfly offline. @rudyfly is right. we would use specific config file when we write daemon related case. If not, the `/etc/pouch/config.json` will impact the running case. 

updated it.

rudyfly:
LGTM

PR #2673, opened 2019-01-16
test: refact travis-ci

rudyfly:
refact travis-ci, make it run faster.

1. remove CRI v1alpha1 test
2. split the integration into multiple parts, let them run together,
because the limit of travis ci jobs, we only can split four parts.
3. update travis-ci os version to ubuntu-16.04 xenial,
because in this version, we can use apt-get to install criu, instead of
building it by ourself, it can reduce the setup time.
4. I have built `nsenter` and upload to aliyun oss storage server,
so only download it instead of building it every time.

NONE

refact test

ci pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
 sudo: required
 
 language: go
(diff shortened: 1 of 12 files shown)


hack/install/install_nsenter.sh
```diff
@@ -29,43 +30,15 @@ nsenter::check_version() {
...
-  tar xf "${tmpdir}/${target}" -C "${tmpdir}"
-
-  cd "${tmpdir}/util-linux-${NSENTER_VERSION}"
-  ./autogen.sh
-  autoreconf -vfi
-  ./configure
-  make
-
-  cp "${cmd}" /usr/local/bin
+  wget "https://${OSS_BUCKET}.${OSS_ENDPOINT}/pouch-test/ubuntu/nsenter-2.24.1" \
```
fuweid: just reminder, if someone wants to run the script without nsenter binary, it might fail. 

however, we can install nsenter manually to let the script work.

travis_ci.sh
```diff
@@ -0,0 +1,182 @@
...
+      run_node_e2e_test
+    ;;
+    *)
+      echo "Unsupport action"
+      exit 1
+    ;;
+  esac
+}
+
+main
```
fuweid: no newline.

travis_ci.sh
```diff
@@ -0,0 +1,182 @@
...
+      -O /usr/local/bin/ossutil
+  sudo chmod +x /usr/local/bin/ossutil
+}
+
+oss() {
+  local action=$1
+  local src=$2
+  local dst=$3
+
+  which ossutil || install_osscmd
```
fuweid: `command -v xxx` is better than `which`

travis_ci.sh
```diff
@@ -0,0 +1,182 @@
...
+  make unit-test
+  make coverage
+
+  bash <(curl -s https://codecov.io/bash) -cF unittest -y .codecov.yml
+}
+
+run_integration_test() {
+  local job_id=$1
+
+  sudo systemctl stop docker
```
fuweid: if there is not docker.service?

.travis.yml
```diff
@@ -22,53 +22,29 @@ notifications:
...
+      script: ./travis_ci.sh integrationtest 4 1
+    - stage: test
+      name: "Integration test 2"
+      script: ./travis_ci.sh integrationtest 4 2
+    - stage: test
+      name: "Integration test 3"
+      script: ./travis_ci.sh integrationtest 4 3
+    - stage: test
+      name: "Integration test 4"
+      script: ./travis_ci.sh integrationtest 4 4
```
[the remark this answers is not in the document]
rudyfly: why need newline at the end?

travis_ci.sh
rudyfly: will be failed.... so I need check it before...

travis_ci.sh
rudyfly: ok

hack/install/install_nsenter.sh
rudyfly: nsenter is public in aliyun oss, so everyone can get it.

.travis.yml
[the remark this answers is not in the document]
rudyfly: Updated

rudyfly:
PTAL @fuweid @allencloud

travis_ci.sh
```diff
@@ -0,0 +1,192 @@
...
+  local index=1
+  local loop=1
+
+  # shellcheck disable=SC2002
+  sum=$(cat "${LOCAL_CASE_FILE}" | wc -l)
+  nums=$((sum / JOBS + 1))
+
+  rm -rf "${LOCAL_CASE_FILE}.*"
+  # shellcheck disable=SC2013
+  for test in $(cat "${LOCAL_CASE_FILE}"); do
```
zhuangqh: using `split` command. `split -l ${nums} -d --suffix-length=1 ${LOCAL_CASE_FILE} ${LOCAL_CASE_FILE}.`

an example 
```bash
[root@pouch test]# ls
example
[root@pouch test]# cat example
1
2
33
44
5
6
7
[root@pouch test]# split -l 3 -d --suffix-length=1 example name.
[root@pouch test]# ls
example  name.0  name.1  name.2
[root@pouch test]# cat name.0
1
2
33
[root@pouch test]# cat name.1
44
5
6
[root@pouch test]# cat name.2
7
```

rudyfly: wow, good idea. thank you for your suggestion.

PR #2675, opened 2019-01-17
bugfix: /etc/mtab will miss quota id

rudyfly:
if set `/etc/mtab` before set rootfs quota, it will make /etc/mtab miss
quota id of rootfs quota, so it need be moved after set rootfs quota.

with overlay work dir, it disappear `work/work`, when only set the dir
`work`'s quota id, if we remove set dir quota id by recursively, the
quota id for overlay is invalid.

NONE

I have no good idea

### DIFF
--- a/storage/quota/quota.go
+++ b/storage/quota/quota.go
@@ -231,6 +231,10 @@ func SetRootfsDiskQuota(basefs, size string, quotaID uint32) (uint32, error) {
 		if err := SetDiskQuota(dir, size, quotaID); err != nil {
 			return 0, errors.Wrapf(err, "failed to set dir(%s) disk quota", dir)
 		}
+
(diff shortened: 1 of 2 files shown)


storage/quota/quota.go
```diff
@@ -231,6 +231,10 @@ func SetRootfsDiskQuota(basefs, size string, quotaID uint32) (uint32, error) {
 		if err := SetDiskQuota(dir, size, quotaID); err != nil {
 			return 0, errors.Wrapf(err, "failed to set dir(%s) disk quota", dir)
 		}
+
+		if err := SetQuotaForDir(dir, quotaID); err != nil {
```
Ace-Tang: Does this operate slow start time when lot of files locates

rudyfly: No, because of overlay Upper and Work have no files.

PR #2678, opened 2019-01-21
bugfix: make sure SnapshotID not empty when start container

HusterWan:

Found a problem on production environment:
```
time="2019-01-21T16:56:13.761316917+08:00" level=error msg="failed to create new containerd container: failed to create container 1edc1d9a416f3ed03d91fd941f040ae16c7a7376c24bf0b5370147c4cc9350e3: snapshot  does not exist: not found: not found"
```

I checked the container's meta.json file and found the `SnapshotID` is empty, though i cannot figure out why this field was empty, but we should make sure the `SnapshotID` should be assigned to proper value.

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -761,10 +761,11 @@ func (mgr *ContainerManager) createContainerdContainer(ctx context.Context, c *C
 		IO:             mgr.IOs.Get(c.ID),
 		RootFSProvided: c.RootFSProvided,
 		BaseFS:         c.BaseFS,
-		SnapshotID:     c.SnapshotID,
 		UseSystemd:     mgr.Config.UseSystemd(),
 	}
 	c.Unlock()
+	// make sure the SnapshotID got a proper value
+	ctrdContainer.SnapshotID = c.SnapshotKey()
 
 	if checkpointID != "" {
 		checkpointDir, err = mgr.getCheckpointDir(c.ID, checkpointDir, checkpointID, false)


fuweid:
@HusterWan the following code is used for old containerd shim right? 

```
// SnapshotKey returns id of container's snapshot
func (c *Container) SnapshotKey() string {
        c.Lock()
        defer c.Unlock()
        // for old container, SnapshotKey equals to Container ID
        if c.SnapshotID == "" {
                return c.ID
        }

        return c.SnapshotID
}
```

HusterWan:
> @HusterWan the following code is used for old containerd shim right?
> 
> ```
> // SnapshotKey returns id of container's snapshot
> func (c *Container) SnapshotKey() string {
>         c.Lock()
>         defer c.Unlock()
>         // for old container, SnapshotKey equals to Container ID
>         if c.SnapshotID == "" {
>                 return c.ID
>         }
> 
>         return c.SnapshotID
> }
> ```

@fuweid No, this `old container` is not `old shim`, but before `upgrade` container

daemon/mgr/container.go
```diff
@@ -760,9 +760,10 @@ func (mgr *ContainerManager) createContainerdContainer(ctx context.Context, c *C
 		IO:             mgr.IOs.Get(c.ID),
 		RootFSProvided: c.RootFSProvided,
 		BaseFS:         c.BaseFS,
-		SnapshotID:     c.SnapshotID,
```
Ace-Tang: I check that snapshot id is first given to `SnapshotID` , ` snapID := id`,  `SnapshotID:    snapID`, so why this field will be empty ?

HusterWan: > I check that snapshot id is first given to SnapshotID , snapID := id, SnapshotID: snapID, so why this field will be empty
I also cannot figure it out why this happened, maybe some extreme cases

>  if this happened, do c.SnapshotKey() will also be empty
No, `SnapshotKey()` will return c.ID if the `SnapshotID` is empty

Ace-Tang:
Look good to me, make the ci pass, @HusterWan

PR #2679, opened 2019-01-21
bugfix: fix containerd pid may be reuse

HusterWan:

In case containerd process may still alive on the host when the pouchd starting, we cache the containerd pid in a file `container.pid` and the pid file will be deleted when the pouchd exited.

In this case, we can check if the pid of containerd still alive, if true pouchd will not run a new containerd instance, but just take over the old containerd instance.

But there has a situation that the pid file cannot be deleted that the host was broken down by accident, and when the host restart again the pid file record a pid reused by the other process but the containerd. So when the pouchd restarting, it will check the containerd is alived but cannot take over the process because the process is not the containerd. 

add PouchDaemonSuite.TestContainerdPIDReuse

### DIFF
--- a/ctrd/supervisord/daemon.go
+++ b/ctrd/supervisord/daemon.go
@@ -167,6 +167,30 @@ func (d *Daemon) healthPostCheck() error {
 	return fmt.Errorf("health post check failed")
 }
 
+func (d *Daemon) isContainerdProcess(pid int) (bool, error) {
+	if !utils.IsProcessAlive(pid) {
+		return false, nil
+	}
+
+	// get process path by pid, if readlink -f command exit code not equals 0,
+	// we can confirm the pid is not owned by containerd.
+	output, err := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
+	if err != nil {
+		logrus.WithField("module", "ctrd").WithField("containerd-pid", pid).Infof("got err: %v", err)
+		return false, nil
+	}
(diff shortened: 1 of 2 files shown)


ctrd/supervisord/daemon.go
```diff
@@ -167,6 +167,29 @@ func (d *Daemon) healthPostCheck() error {
...
 }
 
+func (d *Daemon) isContainerdProcess(pid int) (bool, error) {
+	if !utils.IsProcessAlive(pid) {
+		return false, nil
+	}
+
+	// get process path by pid, if readlink -f command exit code not equals 0,
+	// we can ensure the pid is not owned by containerd.
+	output, err := exec.Command("readlink", "-f", fmt.Sprintf("/proc/%d/exe", pid)).Output()
```
Ace-Tang: Can we use `os.Readlink()`

ctrd/supervisord/daemon.go
```diff
@@ -167,6 +167,29 @@ func (d *Daemon) healthPostCheck() error {
...
+	if err != nil {
+		return false, nil
+	}
+	processPath := strings.TrimSpace(string(output))
+
+	// get containerd process path
+	output, err = exec.Command("which", d.binaryName).Output()
```
Ace-Tang: Use `exec.LookPath()` ?

rudyfly:
LGTM

rudyfly:
LGTM

PR #2682, opened 2019-01-22
bugfix: check old volume mode

rudyfly:
when use volumes-from, need to check old volume mode.

with R/W mode, if hasn't set in new mode, we should inherit from
old mountpoint R/W mode.

with copy mode, if hasn't set `nocopy` in new mode, and also hasn't set
`nocopy` in old mountpoint mode, by default, we should set `CopyData` is
true.

NONE

may be test it manually

1. start container with `copydata` volume, and its image's directory has no data.
2. start new container with `volumes-from`, and the volume directory in new image has data
3. check new container's volume directory whether has image's data.

### DIFF
--- a/daemon/mgr/container_storage.go
+++ b/daemon/mgr/container_storage.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/alibaba/pouch/apis/opts"
 	"github.com/alibaba/pouch/apis/types"
@@ -322,7 +323,6 @@ func (mgr *ContainerManager) getMountPointFromContainers(ctx context.Context, co
 				Destination: oldMountPoint.Destination,
 				Driver:      oldMountPoint.Driver,
 				Named:       oldMountPoint.Named,
-				RW:          oldMountPoint.RW,
 				Mode:        oldMountPoint.Mode,
 				Replace:     oldMountPoint.Replace,
 				Propagation: oldMountPoint.Propagation,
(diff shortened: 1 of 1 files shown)


daemon/mgr/container_storage.go
```diff
@@ -345,8 +345,19 @@ func (mgr *ContainerManager) getMountPointFromContainers(ctx context.Context, co
...
 
-			// the volumes from VolumeFrom is not allowed to CopyData
-			mp.CopyData = false
+			// if mode has no set rw mode, so need to inherit old mountpoint rw mode.
+			if !(strings.Contains(mode, "ro") || strings.Contains(mode, "rw")) {
+				mp.RW = oldMountPoint.RW
+			}
+
+			// if mode has no set copy mode, so need to inherit old mountpoint copy mode.
+			if !strings.Contains(mode, "nocopy") {
```
HusterWan: How about:
```
if !strings.Contains(mode, "nocopy") && !strings.Contains(oldMountPoint.Mode, "nocopy") {
    mp.CopyData = true
}
```

`CopyData` default is `false`

rudyfly: > How about:
> 
> ```
> if !strings.Contains(mode, "nocopy") && !strings.Contains(oldMountPoint.Mode, "nocopy") {
>     mp.CopyData = true
> }
> ```
> `CopyData` default is `false`

Actually, `opts.ParseBindMode` has changed `CopyData` to true.

HusterWan:
LGTM

PR #2686, opened 2019-01-23
test: skip the flay case

fuweid:

When the init process in a container exits, the pouchcontainer daemon
will handle the cleanup job, such as update status code, delete the
containerd's task/container record. The action is handled in the
goroutine. Since pouchcontainer doesn't lock the whole container, when
the status has been updated from `running` into `stopped`, the
containerd's task/container records are not removed by the cleanup job.
if the user wants to use `wait` command to make sure that cleanup job
has been done, it will fail because the status is updated first.

The flaky case is running on this case. We can run it until we have
better solutions to make the status notification more stable.

none

no need


HusterWan:
lgtm

Issue #2685, opened 2019-01-23
[Bug] memory controller missing on ubuntu


HusterWan:
### Ⅰ. Issue Description

We have a test case in: https://github.com/alibaba/pouch/blob/master/test/cli_run_cgroup_test.go#L36

This case is to test `CgroupParent` path, but we got an error on ubuntu:
```
FAIL: cli_run_cgroup_test.go:36: PouchRunCgroupSuite.TestRunWithCgroupParent
cli_run_cgroup_test.go:38:
    // cgroup-parent relative path
    testRunWithCgroupParent(c, "pouch",
        "TestRunWithRelativePathOfCgroupParent")
cli_run_cgroup_test.go:67:
    c.Fatalf("failed to Stat container %s cgroup mountpoint %s: %v", name, file, err)
... Error: failed to Stat container TestRunWithRelativePathOfCgroupParent cgroup mountpoint /sys/fs/cgroup/memory/pouch/eca63851d772b647efb8491a189e7268196c571e7faefef381ce2fa2215ceb5a/memory.limit_in_bytes: stat /sys/fs/cgroup/memory/pouch/eca63851d772b647efb8491a189e7268196c571e7faefef381ce2fa2215ceb5a/memory.limit_in_bytes: no such file or directory
```

This bug may relate to: https://stackoverflow.com/questions/15604055/lxc-cgroup-memory-controller-missing

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1.
2.
3.

### Ⅵ. Environment:

- pouch version (use `pouch version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Ace-Tang:
Not related to the linked issue
```
// GetOwnCgroup returns the relative path to the cgroup docker is running in.
func GetOwnCgroup(subsystem string) (string, error) {
    cgroups, err := ParseCgroupFile("/proc/self/cgroup")
    if err != nil {
        return "", err 
    }   

    return getControllerPath(subsystem, cgroups)
}
```
in ubuntu 16.04, the `/proc/self/cgroup` shows like 
```
12:freezer:/
11:pids:/user.slice/user-1000.slice/user@1000.service
10:perf_event:/
9:devices:/user.slice
8:cpuset:/
7:hugetlb:/
6:rdma:/
5:net_cls,net_prio:/
4:blkio:/user.slice
3:memory:/user.slice
[...]

PR #2693, opened 2019-01-24
test: skip test cgroup parent when use cgroup ns

Ace-Tang:
in ##2688, I make a mistake to find container cgroup path, since when
containeruse cgroup namespace, we can not get cgroup path from container
with`exec cat proc/self/cgroups`. Skip some test if container use cgroup
namespace.

fixes: #2685


rudyfly:
LGTM

Issue #2697, opened 2019-01-28
[bug] pouch inspect -f {{.ExecIDs}} returned error


zjumoon01:
### Ⅰ. Issue Description
pouch inspect -f {{.ExecIDs}} is not supported.

> #pouch inspect -f {{.ExecIDs}} test-lifecycle-exec-01251548410259
>Error: Template parsing error: template: :1:2: executing "" at <.ExecIDs>: can't evaluate field ExecIDs in type *main.inspectContainerJSON

### Ⅱ. Describe what happened
It returned error.

### Ⅲ. Describe what you expected to happen
It should return its ExecIDs.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. start a container
2. pouch exec into the container
3. start a new terminal, and pouch inspect -f {{.ExecIDs}}

### Ⅴ. Anything else we need to know?
In fact, {{.ExecIds}} is supported. But "ExecIds" is a variable in code , not json string.
Pouch should support ExecIDs, not ExecIds 

### Ⅵ. Environment:

- pouch version (use `pouch version`): 1.2.0
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`): 4.9
- Install tools:
- Others:

HusterWan:
You should use `pouch inspect -f {{.ExecIds}}`. But we truly not support inspect `execids` now, I will create PR to fix it.

HusterWan:
assigin @HusterWan

fuweid:
@zjumoon01 could you share your usage here? the execID seems useless thing I guess. Thanks

zjumoon01:
> @zjumoon01 could you share your usage here? the execID seems useless thing I guess. Thanks

It is an ops method. Sometimes "exec" is not exited correctly because some bug on docker environment. So inspecting ExecIDs is useful to debug. It will be useful in pouch debug.

PR #2699, opened 2019-01-28
bugfix: we should get a container's execids from inspect

HusterWan:

Now we cannot get a container's execids from inspect interface, So I opened this pull request tries to fix this problem. Now we can get the execids:
```
#  pouch inspect -f {{.ExecIds}} b63f88
[27573dc84a363456df62ac26a6258536c8c989a87bf9cbbe95c0adf9619f1c36]
```

But this solution has a problem, the execids  of container's memory data or `meta.json` file may be incorrect, because we only update a container's execids when we call `Get` Container API.

So we can ensure the API's data is correct, but the data in memory or file may incorrect, this solution may not be best, we can discuss it. How about extend the interface:
```golang
// before
Get(ctx context.Context, name string) (*Container, error)

// after changed, execids not stored in ContainerJSON, but by api interface
Get(ctx context.Context, name string) (*Container, []string, error)
``` 

fixes: #2697 

add unit test TestSafeMapValues
add an integration test  TestContainerInspectExecIds

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -540,6 +540,24 @@ func (mgr *ContainerManager) Get(ctx context.Context, name string) (*Container, 
 	if err != nil {
 		return nil, err
 	}
+	cID := c.Key()
+
+	// get all execids belongs to this container
+	fn := func(v interface{}) bool {
+		execConfig, ok := v.(*ContainerExecConfig)
+		if !ok || execConfig.ContainerID != cID {
+			return false
+		}
+
+		return true
+	}
+
+	var execIDs []string
+	execProcesses := mgr.ExecProcesses.Values(fn)
+	for k := range execProcesses {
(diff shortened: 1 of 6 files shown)


daemon/mgr/container.go
```diff
@@ -541,6 +541,23 @@ func (mgr *ContainerManager) Get(ctx context.Context, name string) (*Container,
 		return nil, err
 	}
 
+	// get all execids belongs to this container
+	fn := func(v interface{}) bool {
+		execConfig, ok := v.(*ContainerExecConfig)
+		if !ok || execConfig.ContainerID != c.Key() {
```
fuweid: I think we can get the `Key` first and use the value. `c.Key()` will call the `lock` which is pretty annoy. WDYT?

HusterWan: good advise :)

PR #2702, opened 2019-01-29
fix cli annotation flagSet to StringArrayVar

wangforthinker:
flagset.StringSlice will split args by comma so that the annotation parse job fails. Should use StringArray for env/labels.

NONE.

add test.

add test.

related to https://github.com/alibaba/pouch/pull/2171.


wangforthinker:
ping @fuweid

PR #2709, opened 2019-02-13
Testing cve

fuweid:
According to [runc CVE issue](https://github.com/opencontainers/runc/commit/0a8e4117e7f715d5fbeef398405813ce8e88558b#commitcomment-32280015), need to run test cases again.


fuweid:
@Ace-Tang and @allencloud 

The 1.2.0 works for the new runC version with updating the case. I will update the case later.

This PR is used to verify the runC version, no need to merge. Close it.

Issue #2649, opened 2019-01-04
kube-proxy start failed when deploy k8s v1.12.0 (with kubeadm) + pouch 1.1.0


xwisen:
## Question
kube-proxy start failed when deploy k8s v1.12.0 + pouch 1.1.0
### os
```shell
[root@k8s-1 ~]# cat /etc/redhat-release
CentOS Linux release 7.6.1810 (Core)

```
### k8s
```shell
[root@k8s-1 ~]# kubeadm version
kubeadm version: &version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T07:06:49Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
[root@k8s-1 ~]# kubelet --version
Kubernetes v1.12.4

```
### pouch
```shell
[root@k8s-1 ~]# pouch version
KernelVersion:   3.10.0-957.1.3.el7.x86_64
Os:              linux
Version:         1.1.0
APIVersion:      1.24
Arch:            amd64
BuildTime:       2018-12-24T10:07:01+00:00
GitCommit:       1.1.0
GoVersion:       go1.10.4

```
### kube-proxy daemonset yaml
```yaml
[root@k8s-1 ~]# kubectl -n kube-system get daemonset kube-proxy -o yaml
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  creationTimestamp: 2019-01-04T10:03:50Z
  generation: 1
  labels:
    k8s-app: kube-proxy
  name: kube-proxy
  namespace: kube-system
  resourceVersion: "1218"
  selfLink: /apis/extensions/v1beta1/namespaces/kube-system/daemonsets/kube-proxy
  uid: 08c5be85-1008-11e9-8fd9-080027d3613a
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: kube-proxy
  template:
    metadata:
      annotations:
        scheduler.alpha.kubernetes.io/critical-pod: ""
      creationTimestamp: null
      labels:
        k8s-app: kube-proxy
    spec:
      containers:
      - command:
        - /usr/local/bin/kube-proxy
        - --config=/var/lib/kube-proxy/config.conf
        image: k8s.gcr.io/kube-proxy:v1.12.0
[...]

PR #2720, opened 2019-02-21
bugfix: we should make MaskedPaths and ReadonlyPaths to be nil when privileged be set

HusterWan:

When creating a privileged container, we should let the `proc` directories to be writable for the container.

Now we have some users that have kubernetes 1.12/1.13 + pouch environment, but they can not start the kube-proxy daemonset pod because the container do not have write access to the `proc` directories.

BTW, the `MaskedPaths` and `ReadonlyPaths` parameters should belong to `HostConfig` not `ContainerConfig`.

fixes: #2649 

yes add ut for the `MaskedPaths` and `ReadonlyPaths` parse


HusterWan:
Verified on the kubernetes 1.13 , the kube-proxy daemonset has been up normally.

ping @fuweid @rudyfly  , can you please help to review this pr? 😆

daemon/mgr/spec_linux.go
```diff
@@ -69,21 +69,18 @@ func populatePlatform(ctx context.Context, c *Container, specWrapper *SpecWrappe
 	// setup something depend on privileged authority
 	if !c.HostConfig.Privileged {
 		s.Linux.MountLabel = c.MountLabel
+
+		// if MaskedPaths or ReadonlyPaths are set, we will use them, otherwise using the default values.
+		if len(c.HostConfig.MaskedPaths) > 0 {
```
fuweid: check is nil here? because if user passes the empty array here? should we care this?

daemon/mgr/spec_linux.go
```diff
@@ -69,21 +69,18 @@ func populatePlatform(ctx context.Context, c *Container, specWrapper *SpecWrappe
...
+			s.Linux.MaskedPaths = c.HostConfig.MaskedPaths
```
fuweid: how about using copy array here?

daemon/mgr/spec_linux.go
fuweid: both no-pass and pass-empty are the same. synced with @HusterWan

daemon/mgr/spec_linux.go
fuweid: no need to copy here. synced with @HusterWan

daemon/mgr/spec_linux.go
HusterWan: if we do not pass this parameter, we will generate a default value. 
refer: https://github.com/alibaba/pouch/blob/master/oci/spec_default.go#L88

PR #2736, opened 2019-03-07
daemon/mgr/image: update deadline for loading images

fuweid:

PouchContainer will load all the images from containerd into memory. At
the beginning, we assume that it can load it in 10 secs. But if the
system has busy IO, it will take long time to load it, especially the
more-layers and huge-size images. So update it from 10 secs to 10 mins.

none

no need.

CI

### DIFF
--- a/daemon/mgr/image.go
+++ b/daemon/mgr/image.go
@@ -30,7 +30,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var deadlineLoadImagesAtBootup = time.Second * 10
+// The daemon will load all the images from containerd into memory. At
+// the beginning, we assume that it can load it in 10 secs. But if the
+// system has busy IO, it will take long time to load it, especially the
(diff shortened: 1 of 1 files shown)


Ace-Tang:
Wait for ci pass

daemon/mgr/image.go
```diff
@@ -30,7 +30,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var deadlineLoadImagesAtBootup = time.Second * 10
+// NOTE: make it to 10 minutes because it will take long time to load the
```
Ace-Tang: Actually this is not overtime because of image, but the machine env, like load, or other thing, would you mind to change the comment? @fuweid

fuweid: yeah. it is one of reason. the biggest issue is high load in machine.

Ace-Tang: Actually not vary accurately， also ok

PR #2742, opened 2019-03-11
ctrd: ignore not found error in clean process

Ace-Tang:
in remove container, we get a clean process, but it only take effect
when unexpect error happen, so ignore error in delete task if container
is not exist, that means remove container is successful.


HusterWan:
LGTM

rudyfly:
LGTM

PR #2751, opened 2019-03-15
test: pull image before using it

fuweid:

The case maybe run by name. If we don't pull image before, the case will
fail. Each case should do their own setup job.

none

no need

CI


rudyfly:
LGTM

PR #2755, opened 2019-03-17
test: pull image before using it

fuweid:

pull image before using it in test suite

none

no need

CI


fuweid:
> IMO, it's better to add PullImage to SetUpTest, wdyt?

This is case for image list. And I think we should pull image in one case because no case doesn't want it.

ZYecho:
LGTM

PR #2758, opened 2019-03-18
Revert "feature: independent runtime root"

fuweid:
Reverts alibaba/pouch#2748


fuweid:
we need to evaluate the impact to the existing running container. before we have fully tested on this, we revert it first.

PR #2759, opened 2019-03-18
feature: support containerd shim v2

### LINKED ISSUE #2749: [feature request] support containerd v2 shim
## Why you need it?

using custom shim for different runtimes.

After upgrade to containerd v1.2, it's time to introduce this feature to PouchContainer.

## How it could be?

TBD

zhuangqh:

fixes #2749 

- should be compatible with legacy version PouchContainer
- runtime dir just using `/run`


daemon/mgr/container_utils.go
```diff
@@ -105,35 +107,45 @@ func (mgr *ContainerManager) generateName(id string) string {
...
-		rPath = filepath.Join(mgr.Config.HomeDir, RuntimeDir, runtime)
+	var options interface{}
+	switch o := r.Options.(type) {
+	case *runctypes.RuncOptions:
+		options = &runctypes.RuncOptions{
+			Runtime:       r.Path,
+			RuntimeRoot:   ctrd.RuntimeRoot,
+			CriuPath:      o.CriuPath,
+			SystemdCgroup: mgr.Config.UseSystemd(),
+		}
```
zhuangqh: generate a brand new options according to the configuration which were passed from config file, thus, there is no data race.
we directly replace some fields here, because these fields are allowed to be configurable directly.

daemon/daemon_utils.go
```diff
@@ -35,20 +41,50 @@ func initialRuntime(baseDir string, runtimes map[string]types.Runtime) error {
...
+		b, err := json.Marshal(r.Options)
+		if err != nil {
+			return fmt.Errorf("failed to marshal options, runtime: %s: %v", name, err)
+		}
+		options := getRuntimeOptionsType(r.Type)
+		if err := json.Unmarshal(b, options); err != nil {
+			return fmt.Errorf("failed to unmarshal to type %+v: %v", options, err)
 		}
+
+		r.Options = options
```
Ace-Tang: I think this is not correct, runtime type is different from runtime which we add in config. For runtime type, it can be pass by cri , we do not need to add them into config, and validate it.
As we get more runtime concept, we can remove the runtime validate, just add parameter here (it can be disscussed)

daemon/mgr/container.go
```diff
@@ -400,8 +400,9 @@ func (mgr *ContainerManager) Create(ctx context.Context, name string, config *ty
 		config.HostConfig.Runtime = mgr.Config.DefaultRuntime
 	}
 
-	if _, exist := mgr.Config.Runtimes[config.HostConfig.Runtime]; !exist {
-		return nil, errors.Wrapf(errtypes.ErrInvalidParam, "unknown runtime %s", config.HostConfig.Runtime)
+	config.HostConfig.RuntimeType, err = mgr.getRuntimeType(config.HostConfig.Runtime)
+	if err != nil {
```
Ace-Tang: Looks weird to validata runtime type,

daemon/daemon_utils.go
zhuangqh: What passed from CRI is also an alias name of actual runtime type and runtime options.

ContainerManager side should maintain a map (name -> runtime config), and CRI would use the name to choose the actual containerd runtime config.

daemon/mgr/container.go
```diff
@@ -780,7 +781,7 @@ func (mgr *ContainerManager) createContainerdContainer(ctx context.Context, c *C
 	// set container's LogPath
 	mgr.SetContainerLogPath(c)
 
-	runtime, err := mgr.getRuntime(c.HostConfig.Runtime)
+	runtimeOptions, err := mgr.generateRuntimeOptions(c.HostConfig.Runtime)
```
zhuangqh: This is the real entrance to create a container. Two situation to goto this method, first, normally create a container from user interface, second, PouchContainer reload containers at restart time.

Inconvenient to store the runtimeOptions to the disk, containerd required a typed pointer, but we would get a map if we reload the value from disk(under restart and reload situation). Thus, we generate options every time.

daemon/daemon_utils.go
Ace-Tang: This can be decided by containerd, why use a alias name ?

Ace-Tang:
I think pouch should less concern the runtime type information, if it is wrong, containerd will tells.

Ace-Tang:
@zhuangqh , could you describe how cri reslove these options, and how we get this from cri, it will help

zhuangqh:
> @zhuangqh , could you describe how cri reslove these options, and how we get this from cri, it will help

alright. On CRI side, there is only a `runtime_handler` option, each CRI implementation should maintain a map (runtime_handler -> config(runtime_type+runtime_options)) . Generate these two value according to the runtime_handler in CRI and pass these two value to containerd.

PR #2761, opened 2019-03-19
test: increase waiting time for daemon related test

fuweid:

The rich-mode container depends on the systemd to start the stopped
process automatically. But it might take more time to starting. 60s
might be too long but just in case that the system environment might
impact the process.

none

no need

CI


HusterWan:
LGTM

PR #2769, opened 2019-03-25
cri: diable snapshotter usage collector by default

fuweid:

If there is config for pouchd, the default value for command flag won't
work. For this case, we should change Disable to Enable.

none

no need

manually check.


rudyfly:
LGTM

PR #2784, opened 2019-03-29
cgroup: make cgroup-parent absolute path

Ace-Tang:
avoid get cgroup path like /sys/fs/cgroup/memory/system.slice/...
since if cgroup path is relative path, controller path get from
`/proc/self/cgroup`.


HusterWan:
LGTM

PR #2831, opened 2019-05-08
fix setup workDir will change the dir ownership on host

HusterWan:

When setting WorkingDir for a container, we should first decide where should create the directory.
If the WorkingDir is in a volume, we should create the dir in the volume.


rudyfly:
LGTM

PR #2885, opened 2019-06-02
feature: add NetworkMode:default in pouch create

lang710:
Signed-off-by: Lang Chi <21860405@zju.edu.cn>

feature: add NetworkMode:default in pouch create

NONE

add it later.

# pouch run -it --net default busybox


Novicei:
Hi, thank you very much for your submission. But I don't quite understand why an additional default mode is added. You should see in the code that if there is a need for bridge mode, you can actively set it, and if no network mode is set, it will default to bridge. But I don't think it's necessary to have a default mode, as that seems like a lot of effort.

PR #2920, opened 2019-06-20
fix: not copy parent dir when path contains dot

ZYecho:

when path contains dot, such as test/. , we should not copy the test dir into dst, we just need to copy the file& dir under test/ into dst.

None.

added.

wait for CI.

None.

### DIFF
--- a/daemon/mgr/container_copy.go
+++ b/daemon/mgr/container_copy.go
@@ -240,8 +240,9 @@ func (c *Container) getResolvedPath(path string, running bool) (resolvedPath, ab
 	}
 
 	// get the real path on the host
-	resolvedPath = filepath.Join(rootfs, absPath)
-	resolvedPath = filepath.Clean(resolvedPath)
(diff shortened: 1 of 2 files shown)


daemon/mgr/container_copy.go
```diff
@@ -240,9 +240,11 @@ func (c *Container) getResolvedPath(path string, running bool) (resolvedPath, ab
 	}
 
 	// get the real path on the host
-	resolvedPath = filepath.Join(rootfs, absPath)
-	resolvedPath = filepath.Clean(resolvedPath)
+	resolvedPath = rootfs + string(os.PathSeparator) + path
+	cleanedPath := filepath.Clean(resolvedPath)
+	resolvedPath = archive.PreserveTrailingDotOrSeparator(cleanedPath, resolvedPath, os.PathSeparator)
 
+	logrus.Debugf("resolved path: %s", resolvedPath)
```
rudyfly: delete this log？

ZYecho: done

PR #2934, opened 2019-06-27
fix: fix removeImage metric logic

### LINKED ISSUE #2923: [bug] logical problem with the metric 'imageActionCounter' calculation
### Ⅰ. Issue Description
when removing a nonexistent image , should not be recorded on the metric 'imageActionCounter'.
### Ⅱ. Describe what happened
when removing a nonexistent image , the metric 'imageActionCounter' with label 'delete' will add one, but the metric 'ImageSuccessActionsCounter' will not add. This makes it noisy to monitor the success rate of deleting image , which means that the success rate is reduced by deleting the non-existing image, but this is not a real problem.

when removing a nonexistent image, the metric 'imageActionCounter' will not add. The metric is recorded only when it actually performs a delete operation.

1.
2.
3.

### Ⅵ. Environment:

- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

KevinBetterQ:
Signed-off-by: KevinBetterQ <1093850932@qq.com>

fixing the statistical logic of ImageActionsCounter metric when removing image

fixes #2923

added

### DIFF
--- a/apis/server/image_bridge.go
+++ b/apis/server/image_bridge.go
@@ -131,7 +131,6 @@ func (s *Server) removeImage(ctx context.Context, rw http.ResponseWriter, req *h
 
 	label := util_metrics.ActionDeleteLabel
 	defer func(start time.Time) {
-		metrics.ImageActionsCounter.WithLabelValues(label).Inc()
 		metrics.ImageActionsTimer.WithLabelValues(label).Observe(time.Since(start).Seconds())
 	}(time.Now())
 
@@ -153,10 +152,12 @@ func (s *Server) removeImage(ctx context.Context, rw http.ResponseWriter, req *h
 	}
 
 	if err := s.ImageMgr.RemoveImage(ctx, name, isForce); err != nil {
+		metrics.ImageActionsCounter.WithLabelValues(label).Inc()
 		return err
 	}
 
 	metrics.ImageSuccessActionsCounter.WithLabelValues(label).Inc()
+	metrics.ImageActionsCounter.WithLabelValues(label).Inc()
 	rw.WriteHeader(http.StatusNoContent)
 	return nil
 }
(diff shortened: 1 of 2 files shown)


ZYecho:
LGTM, @KevinBetterQ could you make sure is this feature still be wanted?

KevinBetterQ:
> LGTM, @KevinBetterQ could you make sure is this feature still be wanted?

For status monitoring, it is indeed necessary

apis/server/image_bridge.go
```diff
@@ -153,10 +152,12 @@ func (s *Server) removeImage(ctx context.Context, rw http.ResponseWriter, req *h
 	}
 
 	if err := s.ImageMgr.RemoveImage(ctx, name, isForce); err != nil {
+		metrics.ImageActionsCounter.WithLabelValues(label).Inc()
```
ZYecho: if the image was not found, I think it will return err in `s.ImageMgr.GetImage(ctx, name)`, so we can judge the err here and add the counter.

apis/server/image_bridge.go
```diff
@@ -131,7 +131,6 @@ func (s *Server) removeImage(ctx context.Context, rw http.ResponseWriter, req *h
 
 	label := util_metrics.ActionDeleteLabel
 	defer func(start time.Time) {
-		metrics.ImageActionsCounter.WithLabelValues(label).Inc()
```
ZYecho: maybe we could judge the return err here to decide how to handle the counter? if err was not found, not add it?

ZYecho: it's more clear I think

Novicei:
Hi, thank you very much for your submission. I probably understand your logic, which means that the status monitoring in the original defer may lead to unnecessary statistics of the number of actions, because it may not be returned when the image operation is performed. But I think the logic is understandable, when you call the remove operation, it should cause the operation count to be incremented by 1, not what you changed.

PR #2958, opened 2019-08-08
bugfix: remove max-size >=2 condition

ZYecho:

no need this condition， because we can still do logrotate when max-file = 1

NONE

NONE

step1. launch pouchd with max-size and max-file flags

 `pouchd --log-opt max-size=10k`
step2. run a container which could write stdout to its log

 `pouch run docker.io/library/busybox:latest top`
step3. check container log is rotated

none


fuweid:
@ZYecho could you please repush it again?

PR #2980, opened 2019-10-11
Makefile: add commit for the version

fuweid:

Since kubernetes will use version in node list, v1.3.0 is too general
and we cannot tell no thing. Based on this, we can set the last commit
for the version.

none

no need


zhuangqh:
test fail, please repush

PR #3003, opened 2019-12-06
fix return wrong error variable

yuzhiquan:
fix return wrong error variable

### Ⅲ. Why don't you add test cases (unit test/integration test)? 

Yes


fuweid:
@yuzhiquan thanks for sending this pr. We are thinking that both `IsUnpack` and `Unpack` actions are retrying. If those actions fails, we only report the warning message and still return the original error created by `Prepare`.

yuzhiquan:
> @yuzhiquan thanks for sending this pr. We are thinking that both `IsUnpack` and `Unpack` actions are retrying. If those actions fails, we only report the warning message and still return the original error created by `Prepare`.

ok, that's explain it.

PR #3015, opened 2020-03-07
Enhancement debian package manager tweaks

Rajpratik71:
Major Changes No 1 : debian package manager tweaks

By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Major Changes No 2 : added packages apt-utils ca-certificates

Because build is

1.  Slow  because "apt-utils" not installed

2. to avoid build to exits with error without having certificate

fixes NONE

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,9 @@ FROM ubuntu:16.04
 # install wget to download golang source code
 # install git
 RUN apt-get update \
-    && apt-get install -y \
+    && apt-get --no-install-recommends install -y \
+    apt-utils \
+    ca-certificates \
     wget \ 
     git \
     make \


Rajpratik71:
@casparant @jumping @furykerry @king6cong  requesting review

Novicei:
Hi, thank you very much for your submission.It seems that the build time can be reduced from 500s to 200s, but there is a problem that after adding --no-install-recommends, the ca-certificates package must be added. When the user wants to rewrite the dockerfile, it may cause Some inconvenience. And can you tell if apt-util is used a lot? As far as I know, not having apt-util installed doesn't make a difference.

PR #3029, opened 2021-04-09
fix: makes pouch stop command support timeout value 0

staugust:
Make `pouch stop -t 0 ${cid}` behaves as its help message describes, that is to say stop container immediately. 
 See https://github.com/moby/moby/blob/68bec0fcf7a5eeb59c027287d06598098edc9f2c/daemon/stop.go#L71
Here's command to verify it: 
```
cid=$(pouch create -it --net=host --entrypoint=cat ubuntu:bionic)
pouch start ${cid}
time pouch stop -t 0 ${cid}
```
For now, `pouch stop` costs about 10 seconds to stop the container that doesn't response to sig-term.  With this commit, pouch stops container immediately.

### DIFF
--- a/daemon/mgr/container.go
+++ b/daemon/mgr/container.go
@@ -955,10 +955,6 @@ func (mgr *ContainerManager) stop(ctx context.Context, c *Container, timeout int
 		return nil
 	}
 
-	if timeout == 0 {
-		timeout = c.StopTimeout()
-	}
-
 	id := c.ID
 	msg, err := mgr.Client.DestroyContainer(ctx, id, timeout)
 	if err != nil {


daemon/mgr/container.go
```diff
@@ -955,10 +955,6 @@ func (mgr *ContainerManager) stop(ctx context.Context, c *Container, timeout int
 		return nil
 	}
 
-	if timeout == 0 {
```
cxz66666, line 958: You can have a look at `ProbeContainer` function, if timeout is 0, it will never return `errtypes.ErrTimeout` message, but you can see ctrd/containerd.go L455, which only call SIGKILL when get `errtypes.ErrTimeout`. 
In other words, if just delete this code directly, it will never delete those containers which ignore `SIGTERM singal`!

cxz66666:
I already add a [pr](https://github.com/alibaba/pouch/pull/3034) which add `kill` command for pouch, after it merge you can simply use `pouch kill [id]` to stop a container immediately!
So I will close this PR

PR #2950, opened 2019-07-15
test: using config file to run daemon test

zhuangqh:
*background*: some env may depend on some
specific config options.

Signed-off-by: zhuangqh <zhuangqhc@gmail.com>

PR #3036, opened 2022-07-18
[test]: using config file to run daemon test

cxz66666:
similar as https://github.com/alibaba/pouch/pull/2950
*background*: some env may depend on some specific config options.  Use configmap to control some env.


rudyfly:
LGTM

PR #3057, opened 2023-09-11
[Bugfix] fix issue of container escape when pouch cp

Wheat2018:

The PR fixed issue that resolvedPath can escape from container to host by using symlink when pouch cp.


rudyfly:
LGTM

